PR #1, opened 2020-09-18
Google cloud services

loicmathieu:
Add support for some popular Google Cloud services.

This PR provides the following integration with Google Cloud services:

- Authentication via our configuration file and allow to inject a GoogleCredentials CDI bean.
- Storage: allow to inject a Storage CDI bean.
- BigQuery: allow to inject a BigQuery CDI bean.
- PubSub
- Spanner: allow to inject a Spanner CDI bean.
- Firestore: allow to inject a Firestore CDI bean.

These extensions make the libraries works in Quarkus and manage the needed dependencies.

Documentation is still missing but erything works including native.
Some works still needs to be done around gRPC.

There is already an extension proposal for Firebase here https://github.com/quarkusio/quarkus/issues/8671 maybe we can also provide it but it seems very specific (Firebase admin server).

This is a move of the Google Cloud Services extension from Quarkus core repo to the Quarkiverse, see the original PR here: https://github.com/quarkusio/quarkus/pull/9745


bigquery/deployment/pom.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>io.quarkiverse</groupId>
```
gsmet, line 6: I think I would prefer having a subgroup for each repository. Otherwise it's going to be a nightmare. Even for the Hibernate projects, we moved to using a subgroup for each project and this is going to be a lot bigger.

@aloubyansky WDYT?

gsmet (changes requested):
Don't do the change right away but let's discuss this before merging.

bigquery/deployment/pom.xml
aloubyansky: Let's clarify why that would be a nightmare. You want to be able to filter things based on groupIds?

gsmet: You will have a big pile of artifacts in a directory rather than having each project properly separated in a directory. No idea from which project an artifact is.

If we have a subgroup, we will have each project at the root of the groupId and then in a subdirectory all the artifacts of the project.

Also you will entirely avoid conflicts this way, which is nice.

aloubyansky: Ok, that's reasonable.

loicmathieu: Usually, project are compound names and artifact ID are coma separated terms.
So in this case should I use `io.quarkiverse.google.cloud.services` as a group ID (OK, I can use gcp as a single name but I give this as an example).
Looking at what OpenTracing contrib are using, they all use the same groupId that's why we do it this way.

The __folder__ issue doesn't bother me, nobody really browse Maven central as a tree of folder anyway.

The naming conflict is a possible issue. For example another repository may offer a google-cloud-services extension for something else ... This is avoidable with carefull review, anyway when a repository will be created in the org there will be some sort of "review process" to be integrated in the quarkiverse (git org, maven repo, BOM maybe ?).

Anyway, I'm OK for both :)
Just giving some counter arguments.

gsmet: Well, the fact that it doesn't bother YOU is not enough. Question is: will it be annoying for some users? And I'm sure it will be.

Having a ton of random unsorted artifacts at different versions (releases won't be synced) at the root of the groupId is not good. We could have hundreds of artifacts there. Your project is a good example: you have several artifacts in there and that will be the case for most projects.

I would go with:
`io.quarkiverse.googlecloudservices` as both the groupId and the root package.

Having a specific root for both will avoid conflicts and split packages and make things easier if we need to script things.

But let's wait for @aloubyansky 's view on this before changing things.

gsmet: For me the important point is that the release cycles of all these projects will be totally independent.

gsmet: And I think it should be a process when someone wants to add something to quarkiverse.

We affect him a root that is descriptive and unique that should be used for both the groupId and the root package.

loicmathieu: For root package, for the moment I don't update is so it's still `io.quarkus`.
Max said that we also need the capability to easily move an extension from the Quarkiverse to Quarkus if needed.
Using a different package name could be an issue.
On the other hand, I agree that having completely separate namespaces (groupId and root package) could be easier to avoid namming conflict if a lot of extensions ends up in the Quarkiverse.

So, I'm OK for using separate groupId and packages, just tell me ;)

gsmet: Yeah, but I don't think we want everyone to use the `io.quarkus` package.

And we could have conflicts between Quarkiverse and Core.

So I really don't think using `io.quarkus` is a good idea.

aloubyansky: I agree with the `io.quarkiverse.` prefix followed by a project-specific part. In general, I don't see why not allow dots in that project-specific part. To me `io.quarkiverse.google.cloud.services` isn't looking bad. It's even easier to read than a single word.

gsmet: No it's a bad idea to use dots. Dots are used as a structural separator. I'm pretty sure you will end up with `io/quarkiverse/google/cloud/services` on Maven Central and that's definitely not what we want.

I agree it would be better with dashes but unfortunately you can't use them in package names.

So I would use what I suggested given the constraints we have for package names.

aloubyansky: Of course it will be subdirs but why is that bad?

gsmet: Because what you want is a structure such as:
```
io/quarkiverse/<project>/
```

gsmet: Trust me on this, if we don't have a clean structure, it's going to be a major pain in the future when we have 100 projects.

loicmathieu: I agree with @gsmet on this one, groupId is hierarchical, if I use `io.quarkiverse.google.cloud.services` it means it's a services project on a cloud projects on a google projet on the Quarkiverse.

This feels a bit wrong as it's just a GoogleCloudServices project on the Quarkiverse.

aloubyansky: Ok, then we all agree.

loicmathieu: - *groupId*: `io.quarkiverse.googlecloudservices` or I can use `io.quarkiverse.gcp` as gcp stands for Google Cloud Platform ...
- *artifactId*: `google-cloud-services`
- *root package*: same as groupId

aloubyansky: I'd let you pick @loicmathieu the final groupId. The rest looks good to me. Thanks!

gsmet:
Let's make it `io.quarkiverse.googlecloudservices`.

That way, we can easily guess the groupId from the repo name. Might be handy for scripting.

bigquery/deployment/pom.xml
loicmathieu: @gsmet @aloubyansky I rename everything is it OK now ?

aloubyansky (approved):
Great work, @loicmathieu. As always :)

glefloch (commented):
Nice work!

manodupont:
Let's merge ? 👍

danipenaperez:
Great news, good job guys!
Please share link to find the docs for usage, (when available).

PR #2, opened 2020-10-01
Init the credentials with the access token if no service account file is available

sberyozkin:
@loicmathieu Hi Loic, here is the draft PR to support the injected access tokens, but only if the security account file is not available. I just need to make sure that when `quarkus-oidc` is on the classpath, then, with both `IdTokenCredential` and `AccessTokenCredential` being produced, it is the latter which is injected into `TokenCredential`...
Maybe going forward a map of credentials can be introduced, for example, when more than one service is used and one can only work with the service account (`PubSub`), the other one like `BigQuery` should use access token, etc...but this simple PR should do for now.

`quarkus-security` is a very light dependency and it should be OK linking these services with it as it also has other types of credentials which might become useful in time

### DIFF
--- a/common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpConfiguration.java
+++ b/common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpConfiguration.java
@@ -19,4 +19,17 @@ public class GcpConfiguration {
      */
     @ConfigItem
     public Optional<String> serviceAccountLocation;
+
+    /**
+     * Enable Google Cloud access token authentication
+     * For example, the access token which is returned as part of OpenId Connect Authorization Code Flow
+     * may be used to access Google Cloud services on behalf of the authenticated user.
+     *
+     * Note that if a service account location is configured then the access token will be ignored even if this property is
+     * enabled.
+     *
+     * Disable this property if the default Google Cloud authentication is required.
+     */
+    @ConfigItem(defaultValue = "true")
+    public boolean accessTokenEnabled = true;
 }
--- a/common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpCredentialProducer.java
+++ b/common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpCredentialProducer.java
@@ -3,19 +3,36 @@ package io.quarkiverse.googlecloudservices.common;
 import java.io.FileInputStream;
 import java.io.IOException;
 
+import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Default;
+import javax.enterprise.inject.Instance;
 import javax.enterprise.inject.Produces;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import com.google.auth.oauth2.AccessToken;
 import com.google.auth.oauth2.GoogleCredentials;
+
+import io.quarkus.security.credential.Credential;
+import io.quarkus.security.credential.TokenCredential;
+import io.quarkus.security.identity.SecurityIdentity;
 
 @ApplicationScoped
 public class GcpCredentialProducer {
 
     @Inject
     GcpConfiguration gcpConfiguration;
+
+    @Inject
+    Instance<SecurityIdentity> securityIdentity;
+
+    @PostConstruct
+    public void verifySecurityIdentity() {
+        if (securityIdentity.isResolvable() && securityIdentity.isAmbiguous()) {
+            throw new IllegalStateException("Multiple " + SecurityIdentity.class + " beans registered");
+        }
+    }
 
     @Produces
     @Singleton
@@ -24,6 +41,13 @@ public class GcpCredentialProducer {
         if (gcpConfiguration.serviceAccountLocation.isPresent()) {
             try (FileInputStream is = new FileInputStream(gcpConfiguration.serviceAccountLocation.get())) {
                 return GoogleCredentials.fromStream(is);
+            }
+        } else if (gcpConfiguration.accessTokenEnabled && securityIdentity.isResolvable()
+                && !securityIdentity.get().isAnonymous()) {
+            for (Credential cred : securityIdentity.get().getCredentials()) {
+                if (cred instanceof TokenCredential && "bearer".equals(((TokenCredential) cred).getType())) {
+                    return GoogleCredentials.create(new AccessToken(((TokenCredential) cred).getToken(), null));
+                }
             }
         }
 
--- a/README.md
+++ b/README.md
@@ -26,6 +26,10 @@ it depends from where your application runs (inside our outside Google Cloud Pla
 
 The current authentication flow is as follows:
 - Check the `quarkus.google.cloud.service-account-location` property, if it exists, use the service account file from this location.
+- Check the access token returned as part of OpenId Connect Authorization Code Grant response after a user has authenticated with
+  Google OpenId Connect provider (see [Quarkus OpenId Connect for Web Applications](https://quarkus.io/guides/security-openid-connect-web-authentication)).
+  This access token can be used to access Google Services on behalf of the currently authenticated user
+  but will be ignored if the `quarkus.google.cloud.accessTokenEnabled` property is set to `false`.
 - Use `GoogleCredentials.getApplicationDefault()` that will search for credentials in multiple places:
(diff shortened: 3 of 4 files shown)


common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpCredentialProducer.java
```diff
@@ -25,6 +31,8 @@ public GoogleCredentials googleCredential() throws IOException {
             try (FileInputStream is = new FileInputStream(gcpConfiguration.serviceAccountLocation.get())) {
                 return GoogleCredentials.fromStream(is);
             }
+        } else if (tokenCredential.getToken() != null) {
+        	return GoogleCredentials.create(new AccessToken(tokenCredential.getToken(), null));
```
sberyozkin, line 35: This `null` is instead of the `Date` which apparently Google API can use to auto-refresh the token but we don't need, `quarkus-oidc` can handle it itself for the code flow ID token which also auto-refreshes the access token

loicmathieu:
Hi thanks for contributing!
If I understand it, the authentication flow will be: service account file -> token credential -> default (which will rely on the service account env variable, or the gcloud auth configuration).

Service account is server to server authentiation whereas token credential is final user authentication, so we may still be able to override the user authentication if needed. It's good for me.

Maybe we need to be able to disable this kind of authentication in case someone want to use this inside a GCP Function or AppEngine where the default authentication works out of the box, this will allow the user to rely on the default authentication for accessing google services (that will be a service account) and still use user authentication for other part of the application.

sberyozkin:
HI @loicmathieu 
First of all, I updated it a bit to use `Securityidentity` instead - it is more flexible (ID token can be picked up if needed), and in fact it will work even with your OAuth2 extension :-), in principle, AFAIK, Google access tokens are binary tokens, though indeed, the primary case for using this access token is with the OpenId Connect Authorization code flow. I can work on a Qute + BigQuery + quarkus-oidc demo as well later on.

> Maybe we need to be able to disable this kind of authentication in case someone want to use this inside a GCP Function or AppEngine where the default authentication works out of the box, this will allow the user to rely on the default authentication for accessing google services (that will be a service account) and still use user authentication for other part of the application.

Good point. Maybe:

```
   /**
     * Enable Google Cloud access token authentication (
     * For example, the access token which is returned as part of OpenId Connect Authorization Code Flow
     * may be used to access Google Cloud services on behalf of the authenticated user.
     *
     * Note that if a service account location is configured then the access token will be ignored even if this property is enabled.
     *
     * Disable this property if the default Google Cloud authentication is required.
     */
    @ConfigItem(defaultValue=true)
    public boolean accessTokenEnabled = true;
```

Or something similar ? thanks

loicmathieu:
@sberyozkin the proposed configuration item is good for me, the documentation part is really great by the way ;)

There is no automated test so I would need to test this manually, and I don't know how to test the oidc flow.

sberyozkin:
@loicmathieu Thanks, let me update the PR then (we can also disable the property by default - can be easily tweaked).

> There is no automated test so I would need to test this manually, and I don't know how to test the oidc flow.

You can authenticate against Google as shown [here](https://quarkus.io/guides/security-openid-connect-multitenancy#google-openid-provider-configuration). Just remove `.tenant-b.` part and make it a default configuration:
```
quarkus.oidc.auth-server-url=https://accounts.google.com
quarkus.oidc.application-type=web-app
quarkus.oidc.client-id={GOOGLE_CLIENT_ID}
quarkus.oidc.credentials.secret={GOOGLE_CLIENT_SECRET}
quarkus.oidc.token.issuer=https://accounts.google.com
quarkus.oidc.authentication.scopes=email,profile,openid
```

and add the `quarkus-oidc` dependency. So the basic test can just involve, without even the cloud services, [injecting the ID token](https://quarkus.io/guides/security-openid-connect-web-authentication#accessing-id-and-access-tokens) and returning your name to the browser.
Once this is done, you can then try check if the access token returned alongside ID token can be used to access some of these services, here is the instructions I added awhile back in the other project in a BigQuery demo (how to make this access token capable of accessing it on behalf of the user), it may be a bit outdated, but should be also applicable to any other Google service:
```

First, create a Big Query project as described in
https://cloud.google.com/bigquery/sign-up

1. OAuth2 Authorization Code flow.

Create a client id and secret as described in 
https://cloud.google.com/bigquery/authorization#clientsecrets.

Set Redirect URI to 
https://localhost:8080/bigquery
(You'd have JAX-RS `@GET` here, with `@Authenticated`) 

Then start a browser and go to 

https://localhost:8080/bigquery
``` 

and after you login in to Google you should land to this JAX-RS resource which will call BigQuery (or some other service)

Give it a try please when you get some time :-)

sberyozkin:
@loicmathieu thanks, I've updated the PR with the boolean property check, would you like me to open it for Review ?

loicmathieu:
@sberyozkin open it for review when you're done but I will not have time to review it untill mid of next week as I wanted to test it so it may take me some time.

sberyozkin:
@loicmathieu Sure, will do, definitely, it does not have to be rushed, so yeah, lets test it etc first :-)

kostacasa:
Could we detect whether Quarkus is running inside of a managed GCP environment by checking for availability of the  [metadata server](https://cloud.google.com/compute/docs/storing-retrieving-metadata)? I imagine it would always be preferable to utilize this if possible and would reduce the need for a configuration property to control where the access (and possibly ID) tokens are fetched from.

loicmathieu:
@kostacasa when running inside a managed GCP environment, you won't need to setup the serciceAccount, the call to `GoogleCredentials.getApplicationDefault();` will return the default credential so it should be OK.

Your note on getting the project ID from the managed environment is a good remark, can you open an issue for that ?

common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpConfiguration.java
```diff
@@ -19,4 +19,16 @@
...
+     * Enable Google Cloud access token authentication
+     * For example, the access token which is returned as part of OpenId Connect Authorization Code Flow
+     * may be used to access Google Cloud services on behalf of the authenticated user.
+     *
+     * Note that if a service account location is configured then the access token will be ignored even if this property is enabled.
+     *
+     * Disable this property if the default Google Cloud authentication is required.
+     */
+    @ConfigItem(defaultValue= "true")
+    public boolean accessTokenEnabled = true;
```
loicmathieu, line 33: Don't assign it to true as you already use `defaultValue`

loicmathieu (commented):
Can you please add documentation into the main README.md at the root of this repository ?

An integration test would also be great but not now as the `integration-tests` module is not yet a multi-module and I don't want the main integration test to use access token authentication. I'll open a followup PR for this later.

loicmathieu:
I successfully tested it thanks to your detailed information !

As soon as you document it, I'll merge it.

common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpConfiguration.java
sberyozkin: @loicmathieu Sure, it is just in Quarkus OIDC this is the pattern we follow to support when the users creating `OidcTenantConfig` manually in their providers...But yeah, here I'll just remove `true`

sberyozkin:
Hi @loicmathieu 

> I successfully tested it thanks to your detailed information !

Cool, thanks for spending the time on it

> As soon as you document it, I'll merge it.

Sure, I suppose I will add a section not specifically about BigQuery but about linking the user acces tokesn with all (or most) of these services. Then later on there will be an integration test, and may be we can have a joint demo too :-)

sberyozkin:
Hi Loic @loicmathieu I've updated `README`, I can update it further if you'd like, no problems, thanks

loicmathieu:
@sberyozkin the current documentation is OK for me, when publishing guides for the Quarkus website (or whatever will store the Quarkiverse guides) we could have a more complete example.

As soon as you fix the code formatting (yes, code style on Quarkiverse is the same as on Quarkus ; ) I'll merge it.

sberyozkin:
@loicmathieu I had use `Instance<SecurityIdentity>` as it may not always be available; I can tweak the verification code there a bit if you'd like

loicmathieu:
@sberyozkin goods for me.

loicmathieu:
@all-contributors please add @sberyozkin for code

PR #55, opened 2020-11-16
Smallrye Reactive Messaging PubSub

loicmathieu:


jfbenckhuijsen:
@loicmathieu was just looking around at this. Are you still working on this? Else I can have a go at it to hook reactive-messaging up to pubsub.

loicmathieu:
I have an old prototype but as the Pub/Sub connector is not supported at SmallRye side I can't merge it.

PR #58, opened 2020-12-01
Modify Quarkus Storage extension to rely on official GraalVM configs

dzou:
This PR modifies the storage module to rely on the official GCP GraalVM configuration declared here: https://github.com/GoogleCloudPlatform/google-cloud-graalvm-support

If this is satisfactory I can move forward with doing this for Pub/Sub as well which is the other service we believe we have mature support for.

For testing - I manually verified that the integration-test app still works.

cc/ @loicmathieu @meltsufin

### DIFF
--- a/storage/runtime/pom.xml
+++ b/storage/runtime/pom.xml
@@ -21,6 +21,20 @@
         <dependency>
             <groupId>com.google.cloud</groupId>
             <artifactId>google-cloud-storage</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>javax.annotation</groupId>
+                    <artifactId>javax.annotation-api</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.checkerframework</groupId>
+                    <artifactId>checker-qual</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>com.google.cloud</groupId>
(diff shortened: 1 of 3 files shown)


storage/runtime/pom.xml
```diff
@@ -21,20 +21,11 @@
...
-                </exclusion>
-                <exclusion>
-                    <groupId>javax.annotation</groupId>
-                    <artifactId>javax.annotation-api</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>org.checkerframework</groupId>
-                    <artifactId>checker-qual</artifactId>
-                </exclusion>
-            </exclusions>
```
loicmathieu, line 37: Why do you remove these exclusions ?
Quarkus relies on jboss-logmanager so we need to exclude commons-loggin, a "compatibility module" is already added from quarkus core.
For javax-annotation Quarkus already includes the corresponding artifact from the Jakarta namespace.
For the check-qual, it's not used at runtime, so we used to exclude it to lower the size of the final artifact.

loicmathieu (changes requested):
I do love a PR that remove a lot of lines ;)

This looks good to me but I would like to make a release of the library with Quarkus 1.10 before merging this one.

storage/runtime/pom.xml
dzou: Done. Sorry - I originally removed them just out of curiosity to see if the integration tests would still work; should have kept them there.

dzou:
> I do love a PR that remove a lot of lines ;)
> 
> This looks good to me but I would like to make a release of the library with Quarkus 1.10 before merging this one.

Sounds good!

meltsufin (approved):
LGTM

loicmathieu (approved):
LGTM, thanks.

I'll check locally and merged it

loicmathieu:
@all-contributors please add @dzou for code

PR #72, opened 2021-01-08
Rely on reflection configurations from google-cloud-graalvm-support

dzou:
Hey Loic,

We just released version `0.2.0` of [`google-cloud-graalvm-support`](https://github.com/GoogleCloudPlatform/google-cloud-graalvm-support) and added reflection configurations for several new libraries including Spanner, Firestore, BigQuery, Bigtable, etc. The configurations are conditionally added -- so for example, we use the GraalVM SDK to determine whether say Spanner libraries are on the classpath, and only add the Spanner configurations if that is true.

I think our project is now mature enough to start officially supporting the reflection configuration needs for Google libraries as a whole. This PR migrates the quarkus extension to depend on the `google-cloud-graalvm-support` dependency for all the modules.

Testing: I ran the integration test main and was able to verify manually that the endpoints for the six services still work.

Here is the PR for your consideration; let me know if you feel ready to make the transition!

cc/ @meltsufin

### DIFF
--- a/bigquery/deployment/src/main/java/io/quarkiverse/googlecloudservices/bigquery/deployment/BigQueryBuildSteps.java
+++ b/bigquery/deployment/src/main/java/io/quarkiverse/googlecloudservices/bigquery/deployment/BigQueryBuildSteps.java
@@ -1,27 +1,12 @@
 package io.quarkiverse.googlecloudservices.bigquery.deployment;
 
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.Set;
-
-import org.jboss.jandex.ClassInfo;
-import org.jboss.jandex.DotName;
-import org.jboss.jandex.Type;
-
 import io.quarkiverse.googlecloudservices.bigquery.runtime.BigQueryProducer;
 import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
-import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
-import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
-import io.quarkus.deployment.builditem.IndexDependencyBuildItem;
-import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
 
 public class BigQueryBuildSteps {
     private static final String FEATURE = "google-cloud-bigquery";
-
-    private static final DotName DOTNAME_GENERIC_JSON = DotName.createSimple("com.google.api.client.json.GenericJson");
 
     @BuildStep
     public FeatureBuildItem feature() {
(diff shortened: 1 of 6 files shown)


gsmet:
That's very cool.

meltsufin (approved):

loicmathieu:
Thanks a lot ! I'll have a look beginning of next week.

loicmathieu (approved):
Again, a lot of deleted lines, I cloud not be more happy with a PR :)

This looks good, but I prefere to test it locally before merging it.

loicmathieu:
@dzou by the way, for GRPC support I have this substitution: https://github.com/quarkiverse/quarkiverse-google-cloud-services/blob/master/common/grpc/src/main/java/io/quarkiverse/googlecloudservices/common/grpc/runtime/graal/GoogleApiGaxGrpcSubstitutions.java

Is it still needed with your library ?
If so, maybe I can upstream it in your library so other may be able to take advantage of it ?

loicmathieu:
All good, thanks again !

dzou:
> @dzou by the way, for GRPC support I have this substitution: https://github.com/quarkiverse/quarkiverse-google-cloud-services/blob/master/common/grpc/src/main/java/io/quarkiverse/googlecloudservices/common/grpc/runtime/graal/GoogleApiGaxGrpcSubstitutions.java
> 
> Is it still needed with your library ?
> If so, maybe I can upstream it in your library so other may be able to take advantage of it ?

Yeah let me do some more digging about this substitution; maybe we can contribute back to the upstream libraries to get a more permanent fix.

What situation was the substitution intended to fix again -- Was it to support deploying a GraalVM app on AppEngine? I'll try reproducing the error to understand it more.

loicmathieu:
This substitution support using GRPC in a native image not deployed on Google Cloud.
It provides a substitution for `com.google.api.gax.grpc.InstantiatingGrpcChannelProvider` without some part of the code that need a `ComputeEngineCredentials` class that is not present if not running on Google Cloud.

See https://github.com/quarkiverse/quarkiverse-google-cloud-services/blob/6b7be8c11cc5f635ae0a757bee66a9bf755fecd1/common/grpc/src/main/java/io/quarkiverse/googlecloudservices/common/grpc/runtime/graal/GoogleApiGaxGrpcSubstitutions.java#L62

I can send a PR upstream with this substitution or the code can be changed upstream to didn't rely on the presence of the `ComputeEngineCredentials` class.

loicmathieu:
Maybe the best is to open an issue somewhere to start discussing it ?

dzou:
> Maybe the best is to open an issue somewhere to start discussing it ?

Sounds good. I'll create an issue in our repo and let's continue our conversation from there.

dzou:
Followup posted here: https://github.com/GoogleCloudPlatform/google-cloud-graalvm-support/issues/65

So far it seems that our sample code in our repo has been working without this substitution. Maybe we can first figure out what the difference is and how to reproduce an error (with the substitution missing). Maybe we are inadvertently getting the `ComputeEngineCredentials` from somewhere in our samples, not sure.

Issue #75, opened 2021-01-23
Enhancement: Secret Manager


madsop:
On my project, I'm using GCP's Secret Manager for handling keys, passwords and other secrets.

I can't make it work with Quarkus Native, but the code for the client library looks manageable in size.

Would be happy to contribute on this one if needed.

madsop:
Sorry, didn't see that this is addressed in https://github.com/quarkiverse/quarkus-google-cloud-services/pull/3

loicmathieu:
@madsop the openned PR in #3 is not going anywhere, the author didn't reply since 3 month now. I may close it soon.

I'll re-open this issue, maybe other would give feedback on this.

dzou:
Hey just a heads up, I'm interested in contributing a secret manager feature once you submit your PR. Basically to let users specify secrets in their `application.properties` files like this:

```
quarkus.db.password=${sm://my-project/my-secret/version}
```

I'll share a proof of concept in the next couple days.

loicmathieu:
This has been provided via #77 and #78

PR #77, opened 2021-01-25
#75: Straight-forward implementation of GCP Secret Manager support

madsop:
For my project, I needed Secret Manager support for Quarkus native, so I decided to try to figure it out myself.

This is an implementation heavily inspired by the pubsub- and firestore-implementation. It's also an implementation that's very close to the original Google client library in user experience.

So I'm drafting this to hear your thoughts on it. Totally understand if it's too early.

### DIFF
--- a/secret-manager/README.md
+++ b/secret-manager/README.md
@@ -32,7 +32,7 @@ This will add the following to your pom.xml:
 ```
 
 ## Some example
-This is an example fetching a single secret from GCP Secret Manager. The first example is in Java, while the second is the equivalent version in Kotlin.
+This is an example fetching a single secret from GCP Secret Manager.
 
 First, you'll have to create the secret in the GCP Secret Manager, as described in Google's documentation at https://cloud.google.com/secret-manager/docs/creating-and-accessing-secrets.
 
@@ -56,26 +56,6 @@ public class GCPSecretManager {
             AccessSecretVersionResponse response = client.accessSecretVersion(secretVersionName);
             return response.getPayload().getData().toStringUtf8();
         }
-    }
-}
-```
-
-```kotlin
-import com.google.cloud.secretmanager.v1.SecretManagerServiceClient
-import com.google.cloud.secretmanager.v1.SecretVersionName
-import org.eclipse.microprofile.config.inject.ConfigProperty
-import javax.enterprise.context.RequestScoped
-
-@RequestScoped
-class GCPSecretManager {
-
-    @ConfigProperty(name = "secretManagerProjectId", defaultValue = "")
-    lateinit var secretManagerProjectId: String
-    
-    fun getSecretFromSecretManager(secretName: String): String {
-        val secretVersionName = SecretVersionName.of(secretManagerProjectId, secretName, "latest")
-        val client = SecretManagerServiceClient.create()
-        return client.accessSecretVersion(secretVersionName).payload.data.toStringUtf8()
     }
 }
 ```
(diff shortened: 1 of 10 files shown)


loicmathieu:
I added a comment on the other PR to ping the contributor, if he didn't respond we should proceed with this one

loicmathieu:
So, #3 is abandonned and can be used as an inspiration.
I'm OK to proceed with this one, I'll review it as soon as it's ready.

madsop:
@loicmathieu , @dzou : I think this is ready for review now. I've added a readme file. I've also looked through the #3 PR and think that I've addressed the relevant comments from that one as well.

secret-manager/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+```xml
+<dependency>
+    <groupId>io.quarkiverse.googlecloudservices</groupId>
+    <artifactId>quarkus-google-cloud-secret-manager</artifactId>
+    <version>${googleCloudServicesVersion}</version>
+</dependency>
+```
+
+## Some example
```
loicmathieu, line 34: Please provides a Java example as all other example are in Java.

secret-manager/README.md
```diff
@@ -0,0 +1,55 @@
+# Quarkiverse - Google Cloud Services - Secret Manager
+
+This extension allows to inject the Google Cloud Platform Secret Manager client library inside your Quarkus application.
```
loicmathieu, line 3: This is not "inject" but "use" as inject is connoted for CDI

loicmathieu (commented):
Thanks

I add some documentation related comments. 

Can you also add in the documentation the step to create the secret used in your example? 

I wonder how it works regarding authentication? It didn't use the gcp credentials provided  by the common lib.

loicmathieu:
Just saw that auth is enhanced on the PR from @dzou so this part is OK for this PR 👌

madsop:
> Thanks
> 
> I add some documentation related comments.
> 
> Can you also add in the documentation the step to create the secret used in your example?
> 
> I wonder how it works regarding authentication? It didn't use the gcp credentials provided by the common lib.

Thanks for providing feedback! Just pushed an updated version of the readme file.

Also thanks to @dzou for improving the auth part, which I didn't put that much focus on. I've been using the GOOGLE_APPLICATION_CREDENTIALS environment variable locally, so that way it has been working on my machine.

secret-manager/README.md
madsop: Indeed, it's nice to be consistent. However, I think it's nice that the project has an example or two in Kotlin as well, for those (me included) who are using Kotlin with Quarkus native, just to prove that this project also can be helpful in that context. I added both a Java and a Kotlin version in the newly pushed commit, perhaps that's a suitable solution?

loicmathieu: The issue here is consistency as this will be the only guide that incldues Kotlin example.
So, if you plan to contribute Kotlin example for the other guide, I'm OK with it (I'll love to see more Kotlin with Quarkus), otherwise it is better to avoid it.

The readability also is a concern, when using mutiple example on the same thing on different languages I expect the design to only show one language and having way to toggle from one to the other, we didn't have this currently ...

madsop: Yeah, I see the concerns you're pointing out. I guess most people using Kotlin these days also know Java pretty well, and are able to convert a Java example into Kotlin - and if not, IntelliJ does that pretty good as well.

loicmathieu (approved):
LGTM,
Thanks a lot!
I'll merge soon, I want to test it locally before merging it.

loicmathieu:
@all-contributors please add @madsop for code and doc

PR #84, opened 2021-02-16
Add Cloud Platform Auth Scope

dzou:
Add the Cloud Platform Auth Scope to enable access to Cloud APIs via service accounts. Add Secret Manager producer to demonstrate issue.

It seems like when using service accounts to authenticate, you need the extra step of calling the `.createScoped(..)` in certain cases. This "cloud-platform" scope comes from the [service account docs](https://cloud.google.com/compute/docs/access/service-accounts#accesscopesiam). We do this in our [Spring code](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-core/src/main/java/com/google/cloud/spring/core/DefaultCredentialsProvider.java#L95) too.

I'm not sure why you need this when GCP already has the "IAM role" thing, but this was the only way I could get this sample to work. BTW, the service account needs to also have access to the secret via the "secret accessor role":

image

If you remove the scope you will be able to reproduce the error when visiting `localhost:8080/secretmanager`.

### DIFF
--- a/common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpCredentialProducer.java
+++ b/common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpCredentialProducer.java
@@ -21,6 +21,8 @@ import io.quarkus.security.identity.SecurityIdentity;
 @ApplicationScoped
 public class GcpCredentialProducer {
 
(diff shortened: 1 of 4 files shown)


loicmathieu:
Thanks, I tested it with my actual service account and it works.

PR #88, opened 2021-03-15
fix: missing arc dependency in common

loicmathieu:


gastaldi (approved):
LGTM but you can also replace the `quarkus-core` with `quarkus-arc` since it is resolved transitively

loicmathieu:
> you can also replace the quarkus-core with quarkus-arc since it is resolved transitively

Yes, I forget about this!
I'll do it later, I will merge as is to fix the ecosystem CI.

PR #89, opened 2021-03-16
Default project

loicmathieu:
- Falback to default project ID from `ServiceOptions.getDefaultProjectId()` when no projectID is set inside `application.properties`.
- Use the default project ID in the various ITs

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,12 @@ They all share an optional common configuration property to set the project ID:
 quarkus.google.cloud.project-id=<your-project-id>
 ```
 
+If the project ID is not set, the extensions will default to using `ServiceOptions.getDefaultProjectId()` 
+that will use the default project detected via Application Default Credentials.
+
 All these extensions works with applications built as native image executables.
 
-These extension works well within the various Google Cloud Functions extensions available inside Quarkus as they directly authenticate via the built-in credentials.
+These extensions works well within the various Google Cloud Functions extensions available inside Quarkus as they directly authenticate via the built-in credentials.
 
 ## Authenticating to Google Cloud
 
--- a/bigquery/runtime/src/main/java/io/quarkiverse/googlecloudservices/bigquery/runtime/BigQueryProducer.java
+++ b/bigquery/runtime/src/main/java/io/quarkiverse/googlecloudservices/bigquery/runtime/BigQueryProducer.java
@@ -28,6 +28,8 @@ public class BigQueryProducer {
     @Default
     public BigQuery bigQuery() throws IOException {
         return BigQueryOptions.newBuilder().setCredentials(googleCredentials)
-                .setProjectId(gcpConfiguration.projectId.get()).build().getService();
+                .setProjectId(gcpConfiguration.projectId)
+                .build()
+                .getService();
     }
 }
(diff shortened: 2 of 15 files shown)


loicmathieu:
@dzou can you review this one if you have some time ?

dzou:
Sure, I will take a look! Thanks for doing this; sorry I remember I said that I would but didn't get a chance to write the PR.

dzou:
What do you think of setting the default value of [`projectId` in `GcpConfiguration.java`](https://github.com/quarkiverse/quarkus-google-cloud-services/blob/master/common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpConfiguration.java#L15) to `ServiceOptions.getDefaultProject()`? I.e.: 

```
String projectId = ServiceOptions.getDefaultProjectId();
```

This way you can rely on `quarkus.google.cloud.project-id` as the single source of truth; no need to invoke `ServiceOptions.getDefaultProjectId()` elsewhere.

loicmathieu:
> What do you think of setting the default value of projectId in GcpConfiguration.java to ServiceOptions.getDefaultProject()? I.e.:

I think about it then discard it but after a second thought this should be a good idea.
What will be the value if there is no default credential set ? (so if not gcloud auth login has been done on the computer)

dzou:
> > What do you think of setting the default value of projectId in GcpConfiguration.java to ServiceOptions.getDefaultProject()? I.e.:
> 
> I think about it then discard it but after a second thought this should be a good idea.
> What will be the value if there is no default credential set ? (so if not gcloud auth login has been done on the computer)

I see, if the project is unset, then `ServiceOptions.getDefaultProject()` should return `null`. I think this is okay since at that point the client library will throw an error about needing to specify a project ID. (At least this is what appears to be done [in Spring Cloud GCP](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-core/src/main/java/com/google/cloud/spring/core/DefaultGcpProjectIdProvider.java#L36).)

loicmathieu:
@dzou I set the default via a build step to avoid all the optional dance.
This looks better now ;)

integration-tests/main/src/main/java/io/quarkiverse/googlecloudservices/it/BigtableResource.java
```diff
@@ -21,12 +20,10 @@
     private static final String TABLE_ID = "test-table";
     private static final String COLUMN_FAMILY_ID = "test-column-family";
 
-    @ConfigProperty(name = "quarkus.google.cloud.project-id")
-    String projectId;
-
     @PostConstruct
     void initBigtable() throws IOException {
-        try (BigtableTableAdminClient adminClient = BigtableTableAdminClient.create(projectId, INSTANCE_ID)) {
+        try (BigtableTableAdminClient adminClient = BigtableTableAdminClient.create(ServiceOptions.getDefaultProjectId(),
```
dzou, line 25: For the integration tests, would it be preferable to still depend on `quarkus.google.cloud.project-id`? (i.e. would it still fallback to `ServiceOptions.getDefaultProjectId()`?) This may just keep the options open a little more if users wanted to override the default project in the integration tests for whatever reason.

dzou (approved):
Looks good! Just have one question about the integration tests using `ServiceOptions.getDefaultProjectId()` directly vs. using the property.

loicmathieu:
@dzou can you have a final look at it before I merge it ?

dzou (approved):
Nice changes, looks good. Very clean!

PR #92, opened 2021-03-31
Allow Storage Host Configuration with Property

### LINKED ISSUE #91: Provide endpoint-override properties
> Contributions are always welcome, but better create an issue to discuss them prior to any contributions.

I've just found this repository and realized I'd rather use this than creating my producers in every single project again and again.
The feature I miss most currently is to provide a custom endpoint.

`StorageOptions.newBuilder().setCredentials(credentials).setHost(gcpConfiguration.hostOverride.orElse(null)).build().getService();`

The sqs lib has the property `quarkus.sqs.endpoint-override` for that.

Use case: In my tests I start testcontainers which usually expose a locahost:port location. Then I use a custom credential (extends GoogleCredentials implements ServiceAccountSigner) to communicate with the mock server.

On a related note, is there a reason why `GcpCredentialProducer` produces `GoogleCredentials` instead of the more generic `Credentials`? Not a big problem, just requires a non-null returning implementation of `refreshAccessToken` when talking with the mock server, while Credentials can be extended with returning all null values.

PeterUlb:
resolves #91

Sadly restrictToAnnotatedClass is not yet available (higher quarkus version?), if the global nature is a problem we could use QuarkusTestResources. But imo it's fine.

Google is a bit incosistent about the naming.

Client Library | Configure Endpoint
-- | --
Datastore | setHost(hostport)
Firestore | setHost(hostport)
Spanner | setEmulatorHost(hostport)
Pub/Sub | setChannelProvider with forTarget(hostport) / setEndpoint(endpoint)
Bigtable | setChannelProvider with forTarget(hostport) / setEndpoint(endpoint)

So for datastore I now went with `quarkus.google.cloud.storage.host-override`, similar to `quarkus.sqs.endpoint-override`.

Any feedback appreciated.

### DIFF
--- a/integration-tests/main/src/main/java/io/quarkiverse/googlecloudservices/it/StorageResource.java
+++ b/integration-tests/main/src/main/java/io/quarkiverse/googlecloudservices/it/StorageResource.java
@@ -13,7 +13,7 @@ import com.google.cloud.storage.Storage;
 
 @Path("/storage")
 public class StorageResource {
-    private static final String BUCKET = "quarkus-hello";
+    public static final String BUCKET = "quarkus-hello";
 
     @Inject
     Storage storage;
--- a/integration-tests/main/src/test/java/io/quarkiverse/googlecloudservices/it/StorageResourceTest.java
+++ b/integration-tests/main/src/test/java/io/quarkiverse/googlecloudservices/it/StorageResourceTest.java
@@ -1,0 +1,64 @@
+package io.quarkiverse.googlecloudservices.it;
+
+import com.google.cloud.storage.Bucket;
+import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.Storage;
+import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.RestAssured;
+import io.restassured.parsing.Parser;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.utility.DockerImageName;
+
+import javax.inject.Inject;
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+
+import static io.restassured.RestAssured.given;
+
+@QuarkusTest
+public class StorageResourceTest {
+    private static final int PORT = 8089;
+
+    private static GenericContainer<?> GCLOUD_CONTAINER;
+
+    @Inject
+    Storage storage;
+
+    @BeforeAll
+    public static void startGcloudContainer() {
+        GCLOUD_CONTAINER = new GenericContainer<>(DockerImageName.parse("fsouza/fake-gcs-server"))
+                .withExposedPorts(4443)
+                .withCreateContainerCmdModifier(cmd -> {
+                    cmd.withEntrypoint("/bin/fake-gcs-server", "-scheme", "http", "-backend", "memory");
+                });
+        GCLOUD_CONTAINER.setPortBindings(Collections.singletonList(String.format("%d:%d", PORT, 4443)));
+        GCLOUD_CONTAINER.start();
+    }
+
+    @AfterAll
+    public static void stopGcloudContainer() {
+        if (GCLOUD_CONTAINER != null) {
+            GCLOUD_CONTAINER.stop();
+        }
+    }
+
+    @Test
+    public void testStorage() {
+        Bucket bucket = storage.get(StorageResource.BUCKET);
+        if (bucket == null) {
+            bucket = storage.create(BucketInfo.newBuilder(StorageResource.BUCKET).build());
+        }
+        bucket.create("hello.txt", "{\"success\": true}".getBytes(StandardCharsets.UTF_8));
+
+        RestAssured.registerParser("text/plain", Parser.JSON);
+        given()
+                .when().get("/storage")
+                .then()
+                .statusCode(200)
+                .body("success", Matchers.equalTo(true));
+    }
+}
--- a/storage/runtime/src/main/java/io/quarkiverse/googlecloudservices/storage/runtime/StorageConfiguration.java
+++ b/storage/runtime/src/main/java/io/quarkiverse/googlecloudservices/storage/runtime/StorageConfiguration.java
@@ -1,0 +1,17 @@
+package io.quarkiverse.googlecloudservices.storage.runtime;
+
+import java.util.Optional;
+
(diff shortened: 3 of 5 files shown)


loicmathieu:
Thanks for the PR, I'll try to review it soon.

> Sadly restrictToAnnotatedClass is not yet available (higher quarkus version?), if the global nature is a problem we could use QuarkusTestResources. But imo it's fine.

I'll plan to upgrade to Quarkus 1.13 soon, I can do it today so you can rebase on the latest version to use it.

loicmathieu:
@PeterUlb I updated to Quarkus 1.13, you should be able to use `restrictToAnnotatedClass ` now

integration-tests/main/src/test/java/io/quarkiverse/googlecloudservices/it/StorageResourceTest.java
```diff
@@ -0,0 +1,22 @@
...
+import io.quarkiverse.googlecloudservices.it.containers.StorageContainer;
+import io.quarkus.test.common.QuarkusTestResource;
+import io.quarkus.test.junit.QuarkusTest;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+import static io.restassured.RestAssured.given;
+
+@QuarkusTest
+@QuarkusTestResource(value = StorageContainer.class, parallel = true) // use restrictToAnnotatedClass when available
```
loicmathieu, line 12: You can use it now ;)

integration-tests/main/src/test/java/io/quarkiverse/googlecloudservices/it/containers/StorageContainer.java
```diff
@@ -0,0 +1,31 @@
+package io.quarkiverse.googlecloudservices.it.containers;
+
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.utility.DockerImageName;
+
+import java.util.Collections;
+import java.util.Map;
+
+public class StorageContainer implements QuarkusTestResourceLifecycleManager {
```
loicmathieu, line 10: Can you put this inside the test as it is done for firestore/bigtable ?
There is no real advantage of providing a test resource as it is not reused anywhere.

integration-tests/main/src/main/java/io/quarkiverse/googlecloudservices/it/StorageResource.java
```diff
@@ -6,26 +6,25 @@
...
-        storage.create(blobInfo, "test".getBytes());
+        storage.create(BucketInfo.newBuilder(BUCKET).build());
+        BlobInfo blobInfo = BlobInfo.newBuilder(BUCKET, FILE).build();
+        storage.create(blobInfo, "{\"success\": true}".getBytes());
 
         Bucket bucket = storage.get(BUCKET);
-        Blob blob = bucket.get("hello.txt");
+        Blob blob = bucket.get(FILE);
         return new String(blob.getContent());
     }
```
loicmathieu, line 29: Please, keep the test idempotent, I need to be able to call this endpoint multiple time as it is used also during modification of the library on a real projet on the real storage runtime.

I'd rather : keep it as before and upload the file to the bucket from your test, or make it idempotent by deleting the buket at the end of the method.
You can also create a separate method for your test if you want but this will just adds complexity

storage/runtime/src/main/java/io/quarkiverse/googlecloudservices/storage/runtime/StorageConfiguration.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.Optional;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "google.cloud.storage", phase = ConfigPhase.RUN_TIME)
+public class StorageConfiguration {
+    /**
+     * Overrides the default service host.
```
loicmathieu, line 12: Maybe add something to say that it's purpose if for testing only ?

loicmathieu (changes requested):

integration-tests/main/src/test/java/io/quarkiverse/googlecloudservices/it/containers/StorageContainer.java
PeterUlb: I couldn’t find a way to update the properties this way. Any idea how we could dynamically update it within the test itself?

integration-tests/main/src/main/java/io/quarkiverse/googlecloudservices/it/StorageResource.java
PeterUlb: So is it intentional that the retrieved file is a different one from the uploaded one? I’ll see what’s best, since for the upload a fully instantiated storage instance is required. It’s possible to provide a data folder to the mock. 
So the current assumption is that a hello.txt file is already in the bucket when this method is called, correct?

storage/runtime/src/main/java/io/quarkiverse/googlecloudservices/storage/runtime/StorageConfiguration.java
PeterUlb: I followed the description here; https://googleapis.dev/java/google-cloud-clients/0.97.0-alpha/com/google/cloud/ServiceOptions.Builder.html#setHost-java.lang.String-

but a remark that it is commonly used for Testing doesn’t hurt.

integration-tests/main/src/test/java/io/quarkiverse/googlecloudservices/it/containers/StorageContainer.java
loicmathieu: Just use properties prefixed with `%test.` this is the easiest way

integration-tests/main/src/main/java/io/quarkiverse/googlecloudservices/it/StorageResource.java
loicmathieu: Yes, currently I use an existing bucket with an existing file inside my test.
My point here is that I should be able to launch the test or launch the apps and call the endpoint and both should works.
The test should use an embedded storage.
The apps should use a real storage instance from my google cloud project.

loicmathieu (approved):
LGTM, thanks a lot!

I'll open a separate issue for the other services that allow overriding endpoints to track those thanks to the info you provided in your PR

loicmathieu:
@all-contributors please add @PeterUlb for code

PR #94, opened 2021-04-02
Create a BOM to manage dependencies

loicmathieu:


loicmathieu:
@aloubyansky I created a BOM to manage the dependencies. Can you have a look please ? I'm not a Maven expert ;)

I add the BOM in my parent pom but didn't add the Quarkus BOM to my own BOM so that user that will use the Google Cloud Services BOM don't risk any dependency issue. Is it the right thing to do ?

bom/pom.xml
```diff
@@ -0,0 +1,134 @@
...
+        <google-cloud-sdk.version>16.1.0</google-cloud-sdk.version>
+        <google-cloud-graalvm.version>0.3.0</google-cloud-graalvm.version>
+        <!-- The Google Cloud BOM includes the 'android' version of Guava so we need to fix the 'jre' version.
+        Make sure to keep these two libraries compatible. -->
+        <guava.version>30.0-jre</guava.version>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <!-- Google Cloud dependencies, imported as a BOM -->
```
gastaldi, line 23: I am not sure you need to add these dependencies here, as they are resolved transitively, but @aloubyansky might know better. The BOM should contain only your extension´s runtime and deployment modules AFAIK

loicmathieu: So you suggest to keep them inside my parent pom ?

The Quarkus BOM contains all dependencies that's why I add the managed version here.

gastaldi: > So you suggest to keep them inside my parent pom ?
Yes, I think so.

> The Quarkus BOM contains all dependencies that's why I add the managed version here.
Gotcha, I think once the BOM is flattened using @aloubyansky´s plugin the result will be the same, but let´s wait for his take on this

aloubyansky: Flattening does not introduce new dependency constraints. It flattens the parent-child hierarchy.
@loicmathieu your bom is supposed to include the artifacts your project produces plus dependency constraints for other artifacts that your artifacts require to function properly.

bom/pom.xml
```diff
@@ -0,0 +1,134 @@
...
+                <artifactId>libraries-bom</artifactId>
+                <version>${google-cloud-sdk.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
+                <groupId>com.google.guava</groupId>
+                <artifactId>guava</artifactId>
+                <version>${guava.version}</version>
+            </dependency>
```
aloubyansky, line 36: This guava constraint after the libraries-bom won't have any effect. If you want to override what the libraries-bom contains, it should be moved before its import. But given that libraries-bom imports guava-bom, it's better to import the whole guava bom. I am not sure though this is necessary at all though.

aloubyansky: Ah, it's the android version that the libraries-bom imports, sorry. It's ok then.

aloubyansky (approved):
It looks good to me. Luckily quarkus-bom imports the same version of the protobuf. Keep in mind that in the platform, quarkus-bom will dominate. If you think this could cause issues regularly, please let us know asap. Hopefully, not. Thanks!

aloubyansky:
@loicmathieu as soon as you release the BOM, we can check how the platform affects your dependencies, even if your tests aren't available yet for integration. So, let me know when it's available in Maven.

loicmathieu:
@aloubyansky 

>  Luckily quarkus-bom imports the same version of the protobuf

I'll check, maybe I don't need to manage these dependencies here, I rely on quarkus-grpc now and it should be transitively managed.

> as soon as you release the BOM, we can check how the platform affects your dependencies

I plan to make a release soon with the BOM, the migration to Quarkus 1.13 and some small fix/enhancements

loicmathieu:
@aloubyansky I didn't manage directly the protobuf version, they comes from the Google Library BOM.

We should investiguate the usage of the Google Library BOM inside the Quarkus BOM, I'll merge this one and open an issue on the Quarkus repo to discuss this.

loicmathieu:
@aloubyansky now I have an issue with the formatter plugin and I have no idea why

```
No plugin found for prefix 'formatter' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/home/runner/.m2/repository), central (https://repo.maven.apache.org/maven2)] -> [Help 1]
```

loicmathieu:
Found why the issue with the formatter plugin arise.
My BOM didn't use the parent, so I need to add the plugin explicitly.

aloubyansky:
@loicmathieu let me know when the BOM is released

loicmathieu:
@aloubyansky 0.6.0 has been released with the new BOM.

aloubyansky:
Congratulations @loicmathieu! The BOM has zero conflicts and can be integrated w/o adjustments (by the platform BOM generator) into the current platform. I am not yet sure, whether the addition of your BOM into the mix has affected other platform members though.
The next step for you is to publish some tests as Maven test-jar artifacts, so they can be integrated into the platform testsuite.

loicmathieu:
@aloubyansky can you please open an issue in this repo with a quick explaintation of what is expected (or an example) ?

aloubyansky:
Sorry, there are actually some adjustments to your BOM once I add it to the platform. See the screenshot attached. I was using my current working branch of the bom generator originally which had a bug.
Screenshot from 2021-04-08 17-46-23

PR #137, opened 2021-06-21
Publish main tests as test jar

### LINKED ISSUE #104: Publish test artifacts to be integrated into the Quarkus platform testsuite
To be able to integrate the Google Cloud Services Quarkus extensions into the Quarkus community platform https://github.com/quarkusio/quarkus-platform, besides the BOM, which already exists, we need to integrate at least some tests that would make sure that the GCS extensions are compatible with the generated platform BOM.

Usually, you would pick some (or even all, depending on the case) of the integration tests from your repository and publish them as `test-jar`s along with your project release. To attach your test artifacts to the project, you should add the following plugin config to the test modules that should be integrated into the platform testsuite.

````
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-jar-plugin</artifactId>
                <executions>
                    <execution>
                        <goals>
                            <goal>test-jar</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
        </plugins>
````

### PR DESCRIPTION
**Pull request by loicmathieu**
Fixes #104

/cc @aloubyansky 

I also remove the skip release flag as I assume I will need to release the test jar.

### DIFF
--- a/integration-tests/main/pom.xml
+++ b/integration-tests/main/pom.xml
@@ -12,8 +12,6 @@
   <name>Quarkus - Google Cloud Services - Integration Tests - Main</name>
 
   <properties>
-    <maven.deploy.skip>true</maven.deploy.skip>
-    <gpg.skip>true</gpg.skip>
     <native.surefire.skip>${skipTests}</native.surefire.skip>
   </properties>
 
@@ -110,6 +108,17 @@
             <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
           </systemProperties>
         </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <executions>
+          <execution>
+            <goals>
+              <goal>test-jar</goal>
+            </goals>
+          </execution>
+        </executions>
       </plugin>
     </plugins>
   </build>

### CODE REVIEW

File: integration-tests/main/pom.xml [hunk 1]
```diff
@@ -12,7 +12,6 @@
   <name>Quarkus - Google Cloud Services - Integration Tests - Main</name>
 
   <properties>
-    <maven.deploy.skip>true</maven.deploy.skip>
     <gpg.skip>true</gpg.skip>
```
aloubyansky, line 15: Isn't gpg also related?

loicmathieu: Oh yes, you should be right.

aloubyansky:
Looks good @loicmathieu. If you want to try integration into the platform before you release, you could add your config similar to e.g. https://github.com/quarkusio/quarkus-platform/blob/main/pom.xml#L190 (make sure this value is set to true https://github.com/quarkusio/quarkus-platform/blob/main/pom.xml#L193)
The first build will take awhile but subsequent builds should be ~15 sec.

loicmathieu:
@aloubyansky thanks, I didn't thought of trying locally the platform. I'll do it before pushing a PR to it to include the library as a member.

Anyway, I'll merge this one as I'll need a release to open a PR to the platform. I'll publish a 2.0.0.RC1 soon.

PR #177, opened 2021-08-24
fix: make the config holder unremovable

loicmathieu:


gsmet:
I think we should try to understand why it is suddenly removed by ArC.

Issue #224, opened 2021-12-31
Adding GCP operations logging exception


Fungrim:
I created a fork and played around with adding support for logging directly to GCP operations. It's in a working proof of concept stage at the moment, and I though I should ask you guys if you think it is a good idea before I add more features and start polishing: https://github.com/Fungrim/quarkus-google-cloud-services/tree/main/logging

The ideas is:

* Log directly to GCP using the Google Java library
* Support both structured and unstructured logging
* The log record `instant` and `level` should be first class citizen regardless
* Support labels for things like POD name, app name etc (GCP logging labels acts as "dimensions" when analyzing logs) 
* Support trace ID as a first class citizen (to be able to link GCP traces with logs)
* Default to ECS format, but make it customizable

The current status is: it works, has an integration test that needs a GCP project and thus isn't in `main`, and a rough, untested version of the ECS format is in. Has a list of todo's, mostly round configuration. Need more unit tests and polishing.

What do you think? Would you like this to end up as a pull request?

loicmathieu:
Hi, please open a PR so we can discuss this.
The integration test can be in the main integration test so I can run it manually. Add a REST resource class with exposes the multiple way to log but don't add any test classes if it could not be run from the CI.

Fungrim:
@loicmathieu Done: #225

loicmathieu:
Let's keep it openned untill the PR is accepted ;)

loicmathieu:
Fixed by #386

PR #225, opened 2022-01-03
Add GCP operations logging

Fungrim:
This PR suggests a new module which adds direct integration with GCP Operations logging using the GCP client library directly in order to support tracing, labeling, etc.

Copy 'n pasted from #224: 

The ideas is:

* Log directly to GCP using the Google Java library
* Support both structured and unstructured logging
* The log record instant and level should be first class citizen regardless
* Support labels for things like POD name, app name etc (GCP logging labels acts as "dimensions" when analyzing logs)
* Support trace ID as a first class citizen (to be able to link GCP traces with logs)
* Default to ECS format for structure logging, but make it customizable


logging/deployment/src/main/java/io/quarkiverse/googlecloudservices/logging/deployment/LoggingBuildSteps.java
```diff
@@ -0,0 +1,59 @@
...
+                JsonFormatter.class,
+                LabelExtractor.class,
+                LoggingConfiguration.class,
+                TraceInfoExtractor.class,
+                TraceInfo.class,
+                JsonHandler.class,
+                TextHandler.class,
+                SimpleFormatter.class,
+                StackTraceArrayRenderer.class);
+    }
```
loicmathieu, line 52: I don't understand why you make them unremovable as they are not CDI beans.

loicmathieu (changes requested):
I didn't make a full review yet but have already some comment.

Using the extension, one should be able to directly use `Logging` instead of a Logger from jboss-logging or SLF4J by injecting a `Logging` object into a class and directly use it (like in https://github.com/googleapis/java-logging#about-cloud-logging). 
Can you add this in the integration test and document it ?

integration-tests/logging/src/main/java/io/quarkiverse/googlecloudservices/it/LoggingResource.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Path("/logging")
+public class LoggingResource {
```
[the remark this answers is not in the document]
Fungrim: Right. Will do.

logging/deployment/src/main/java/io/quarkiverse/googlecloudservices/logging/deployment/LoggingBuildSteps.java
```diff
@@ -0,0 +1,59 @@
...
+    @BuildStep
+    public FeatureBuildItem feature() {
+        return new FeatureBuildItem(FEATURE);
+    }
+
+    @BuildStep
+    public UnremovableBeanBuildItem credentials() {
+        return UnremovableBeanBuildItem.beanTypes(GoogleCredentials.class);
+    }
+
```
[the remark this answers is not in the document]
Fungrim: I added it because it was removed at some point. But that might have been my local setup (same with the other unremovable beans). Will test again without it.

logging/deployment/src/main/java/io/quarkiverse/googlecloudservices/logging/deployment/LoggingBuildSteps.java
Fungrim: I might have had problems with my local setup, because I did run into missing classes during testing. Will test again without them and see what happens.

Fungrim:
@loicmathieu Good point, I'll add an injected logger to the test as well.

integration-tests/logging/src/main/java/io/quarkiverse/googlecloudservices/it/LoggingResource.java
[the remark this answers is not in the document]
Fungrim: Hm, it works perfectly in a separate module. But when I move it into `main` it starts failing _silently_. It logs correctly to console but nothing appears to be hitting GCP. Exciting.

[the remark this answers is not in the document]
Fungrim: I've added eclusions for `grpc-netty-shaded` and `animal-sniffer-annotations` and depended on the same common libs as `PubSub`. But still nothing and `Logging.flush` hangs, so I'm guessing it's something around gRPC despite that.

logging/deployment/src/main/java/io/quarkiverse/googlecloudservices/logging/deployment/LoggingBuildSteps.java
Fungrim: I've minimized it, but the interfaces the can be bound to the CDI context to customize the logging needed to stay (as did the configuration for some reason).

Fungrim:
PR is updated, but as [noted above](https://github.com/quarkiverse/quarkus-google-cloud-services/pull/225#discussion_r778335494) the `itest` fails silently. I have no idea why, but since `Logging.flush` hangs I'm assuming it has something to do with the gRPC or Netty libs.

integration-tests/logging/src/main/java/io/quarkiverse/googlecloudservices/it/LoggingResource.java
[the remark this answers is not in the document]
loicmathieu: If it has something to do with gRPC try including the `io.quarkiverse.googlecloudservices:quarkus-google-cloud-common-grpc` dependency, it should make gRPC works for Quarkus.

integration-tests/main/src/main/resources/application.properties
```diff
@@ -3,7 +3,7 @@
 #quarkus.google.cloud.project-id=
 
 # We use a dummy test projet id in test for the emulators to work
-%test.quarkus.google.cloud.project-id=test-project
+%test.quarkus.google.cloud.project-id=larsan-net
```
loicmathieu, line 6: Please don't change it ;)

Fungrim:
@loicmathieu Sorry about the long silence. But I've managed to have a new look now and figured out why no logs where uploaded: In the logging runtime I'm injecting the Google credentials (in `LoggingProducer`), however the `main` integration test is using mocked credentials from `ApplicationProducerMock` and thus no logs are sent to Google Operations. 

(I had looked at the secret manager implementation, but failed to realize the integration test for secret manager uses a mocked client as well).

loicmathieu:
Yes, I use the main IT in two ways:
- Manual testing against a real GCP project, I usually launch the project using `quarkus:dev` after updating the config to use my project ID and a service account then do curl to manually test (using [GoogleServicesResourcesTest.java](https://github.com/quarkiverse/quarkus-google-cloud-services/blob/main/integration-tests/main/src/test/java/io/quarkiverse/googlecloudservices/it/GoogleServicesResourcesTest.java) should also do the trick).
- Automated testing via the other test cases inside [the it test package](https://github.com/quarkiverse/quarkus-google-cloud-services/blob/main/integration-tests/main/src/test/java/io/quarkiverse/googlecloudservices/it) 

For GCP logging, you should manually test it (I'll also do it before merging your branch) and add a test method inside the [GoogleServicesResourcesTest.java](https://github.com/quarkiverse/quarkus-google-cloud-services/blob/main/integration-tests/main/src/test/java/io/quarkiverse/googlecloudservices/it/GoogleServicesResourcesTest.java).

Fungrim:
Right. I'll do that. However, I also fetched/merged upstream to `1.1.0-SNAPSHOT` and that broke JUL and JBoss logging. Curiously, injecting a `Slf4j` logger still works (as it hits the `QuarkusDelayedHandler`) so something has changed in Quarkus. Let me know if you want to me to close this PR until I've figured it out or if we should keep it open.

loicmathieu:
You can keep it open, if I have time I'll have a look next week.

The change was from Quarkus 2.5 to Quarkus 2.7, something must have disturbed the loggin system between the two. Maybe JNDI that is now disabled by default.

Fungrim:
Thanks for the hint. I'll dig around a bit and see what I can find. I'll ping you when I've fixed it - apart from that I'm pretty happy with the state of the code at the moment.

Fungrim:
I _miiiight_ have gotten a bit carried away with features :-D

Fungrim:
I have good new and bad news. The good one is that the extension seems to work as it should now. The bad? I have no idea why. I was in the middle of digging through the class path and attempting to downgrade `jboss-logging` (which had been updated in Quarkus 2.7) when... er... it magically started working. I have no exact explanation, but it might be that I'd forgotten to do a clean/install on the entire project between versions. I'm assuming there's at least a possibility that there could have been class version errors in pre-compiled code that caused it, but I'm really just guessing.   Embarrassing, but, hey...

So, it looks good now. Going through it again there's one class - `EscJsonFormat` - that is a bit gnarly and completely untested. I'll see if I can sit down and do that this weekend. Otherwise I'm feeling pretty good about it now.

logging/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+* `%test.quarkus.google.cloud.logging.enabled=false`
+
+### Formatting
+This logging system can either use a standard text based format, Elastic Common Schema, or a completely custom schema.    
+
+* `quarkus.google.cloud.logging.format=[TEXT|JSON]`
+
+#### Structured Logging
+By setting the logging `format` to `JSON` this library will default to ECS format. As loca record parameters and MDC values are not
```
loicmathieu, line 87:
```suggestion
By setting the logging `format` to `JSON` this library will default to ECS format. As local record parameters and MDC values are not
```

logging/README.md
```diff
@@ -0,0 +1,228 @@
...
+official part of the ECS specification, these default to `parameters`  and `mdc` respectively, but this can be configured (see below).
+
+* `quarkus.google.cloud.logging.structured.stack-trace.included=[true|false]` - Should stack traces be included? Default to `true`
+* `quarkus.google.cloud.logging.structured.stack-trace.rendering=[array|string]` - Render stack traces as a strings or an arrays of objects, defaults to `string`
```
loicmathieu, line 91:
```suggestion
* `quarkus.google.cloud.logging.structured.stack-trace.rendering=[array|string]` - Render stack traces as a strings or an array of objects, defaults to `string`
```

logging/README.md
```diff
@@ -0,0 +1,228 @@
...
+        if (record.getParameters() == null) {
+            return null;
+        } else {
+            // do something smart here
+        }
+    }
+}
+```
+
+### Tracing
```
loicmathieu, line 174: Should this be replaced by a support for OpenTracing and OpenTelemetry ?

logging/README.md
```diff
@@ -0,0 +1,228 @@
...
+configurable: 
+
+* `quarkus.google.cloud.logging.gcp-tracing.enabled=[true|false]`
+
+# Injecting GCP Logging
+You can inject a `Logging` instance directly. If you do, the configuration for the project to use,
+still apply. 
+
+If you want the configuration for default log, resource type and labels, and default labels, you can
+inject a `WriteOptionsHolder` which contains an array of default write options as configured.
```
loicmathieu, line 208: I don't really understand the WriteOptions stuff, can you explain it to me ?

logging/README.md
```diff
@@ -0,0 +1,228 @@
...
+* `quarkus.google.cloud.logging.structured.stack-trace.included=[true|false]` - Should stack traces be included? Default to `true`
+* `quarkus.google.cloud.logging.structured.stack-trace.rendering=[array|string]` - Render stack traces as a strings or an arrays of objects, defaults to `string`
+* `quarkus.google.cloud.logging.structured.stack-trace.element-rendering=[string|object]` - How to render each stack trace element, defaults to `string`
+* `quarkus.google.cloud.logging.structured.parameters.included=[true|false]` - Should log record parameters be included? Defaults to `true`
+* `quarkus.google.cloud.logging.structured.parameters.field-name` - This is the field name to use in ESC for parameters, defaults to `parameters`
+* `quarkus.google.cloud.logging.structured.mdc.included=[true|false]` - Should the MDC values be included? Defaults to `true`
+* `quarkus.google.cloud.logging.structured.mdc.field-name` - This is the field name to use in ESC for MDC values, defaults to `mdc`
+
+You can use a completely custom structured format by binding a `JsonFormatter` to the CDI context. For example, if you want to filter the 
+included parameters, you could do this:
```
loicmathieu, line 99: Is using a custom JSON formatter really useful and something someone will do ?
If it's a very rare use case, I'll suggest to drop it as most user will use the EcsJson formatteur and there is already a lot of custimization possible.

logging/runtime/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+            <artifactId>quarkus-google-cloud-common</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.googlecloudservices</groupId>
+            <artifactId>quarkus-google-cloud-common-grpc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-grpc-common</artifactId>
+        </dependency>
```
loicmathieu, line 27: Should not be needed as you already include `quarkus-google-cloud-common-grpc`.

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
```diff
@@ -0,0 +1,220 @@
...
+     * Enable or disable the Goocle Cloud logging.
+     */
+    @ConfigItem(defaultValue = "true")
+    public boolean enabled;
+
+    /**
+     * Configure base formatting to be either plain text or
+     * structured json.
+     */
+    @ConfigItem(defaultValue = "TEXT")
```
loicmathieu, line 33: I think there is a way to specify the list of allowed values.
Or maybe use an enum

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     * Configure default labels.
+     */
+    @ConfigItem
+    public Map<String, String> defaultLabel;
+
+    /**
+     * Configured the monitored resource. Please consult the Google
+     * documentation for the correct values: https://cloud.google.com/logging/docs/api/v2/resource-list#resource-types
```
loicmathieu, line 56: Use `@see` javadoc snippet for clickable links

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
```diff
@@ -0,0 +1,220 @@
...
+
+    public enum StackElementRendering {
+        STRING,
+        OBJECT
+    }
+
+    public enum LogFormat {
+        TEXT,
+        JSON
+    }
```
loicmathieu, line 219: I think you can use these enums inside the config class itself, you can check on the core logging extension for example

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/util/StackTraceArrayRenderer.java
```diff
@@ -0,0 +1,123 @@
...
+
+import io.quarkiverse.googlecloudservices.logging.runtime.LoggingConfiguration.StackElementRendering;
+
+/**
+ * This is a utility class for rendering a stack trace as an array
+ * of thrown exceptions. It uses the same calculation as {@link Throwable}
+ * for truncating stacks for causes - i.e. not show the same stack frames when
+ * they are common - and checks for self referencing and circular references.
+ */
+public class StackTraceArrayRenderer {
```
loicmathieu, line 21: Is there really a use case for adding the stack trace as an array of frames instead a single string ?
If it's a rare use case, I prefere to avoid addind it.

README.md
```diff
@@ -98,6 +99,7 @@ Example applications can be found inside the integration-test folder:
 - [main](integration-tests/main): RESTEasy endpoints using all the Google Cloud Services extensions, to be deployed as a standalone JAR.
 - [google-cloud-functions](integration-tests/google-cloud-functions): A Google Cloud HTTP function using Google Cloud Storage. 
 - [app-engine](integration-tests/app-engine): A RESTEasy endpoint using Google Cloud Storage, to be deployed inside Google App Engine.
+- [logging](integration-tests/logging): A single RESTEasy which demonstrates a custom JSON formatter and trace and label extraction.
```
loicmathieu, line 102: This integration test didn't exist anymore.

loicmathieu (changes requested):
Hi, again, thanks for the amazing job ;)

I did a first review (not yet complete), gives some suggestions, and ask some questions.

I think the set of features is too high for a first implementation, if you see the correponding logging-json extension you'll see it has less features while being out for a while: https://github.com/quarkiverse/quarkus-logging-json

As a gentle advise, you should first implement a basic set of features or discuss with the maintainer before sending a PR with this amount of features, now that they are built I'm encline to accept them but each feature has a maintainance cost. 

That being said, your PR is more than welcome and I'm sure we can balance the feature set and maintenance cost without lossing too much of the great investment you take to implement this.

logging/README.md
Fungrim: Not really, it would _complement_ OT. The thing we're doing here is to provide the `traceId` as a first class member to Operations Logging (OL). That way, if you use OT with export to Operations Tracing, you can supply the `traceId` here and operations will connect the logs with the trace. In a future version, we could try to do it automagically, but that would require a dependency on OT - so I decided to let the implementer provide the `traceId` manually.

Fungrim: (Incidentially, I have the code to integrate OpenTelemetry with Operations Tracing, including picking up the existing `traceId` from the Google load balancers. That's a potential future PR if you're interested.)

logging/README.md
Fungrim: 1 - If you want to inject a `com.google.cloud.logging.Logging` to use directly (instead of logging normally to `jboss` or `slf4j` etc), you will need the configured write options. 

2 - The write options are an array, so for simplicity of injection I opted to wrap the array in a provider, and opted to call it a `*Holder` instead since to align the name with `GcpConfigHolder gcpConfigHolder`

logging/README.md
Fungrim: Good suggestion, will fix.

logging/runtime/pom.xml
Fungrim: Check.

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
Fungrim: I'll have a look.

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
Fungrim: Check.

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/util/StackTraceArrayRenderer.java
Fungrim: As I said, I got a bit carried away :-D I'll stash it and we can revive it later if someone asks for it.

Fungrim:
@loicmathieu Thanks for the review! I appreciate your comment about the feature set, and I certainly acknowledge that smaller PR with a prior discussion would have been better. 

As for the selection of the features, here's some comments that might clarify why it ended up as a fairly big chuck of code: I had two concrete goals when I started (based on the work I'm doing professionally with GCP): integrate tracing with with the logs in GCP Operations - this meant that _only_ formatting was not an option I needed to go via the GCP logger itself. And also I wanted structured logging - this could have been accomplished with `logging-json`, but since I wanted tracing it would mean a round-trip from JSON to string and back again (since the GCP logger works on a Map of objects) which I didn't like. ECS seemed simple enough to implement. 

(Then having started I realized the ability to configure/specify labels is a rather big USP when integrating directly with the GCP logger - so I added that, the rest is history).

logging/runtime/pom.xml
Fungrim: Actually, I get this if I try to remove it: 

```
[ERROR] Quarkus Extension Dependency Verification Error
[ERROR] The deployment artifact io.quarkiverse.googlecloudservices:quarkus-google-cloud-logging-deployment::jar depends on the following Quarkus extension deployment artifacts whose corresponding runtime artifacts were not found among the dependencies of io.quarkiverse.googlecloudservices:quarkus-google-cloud-logging:jar:1.1.0-SNAPSHOT:
[ERROR]  io.quarkiverse.googlecloudservices:quarkus-google-cloud-logging-deployment:jar:1.1.0-SNAPSHOT
[ERROR] *  io.quarkus:quarkus-grpc-common-deployment:jar:2.7.0.Fina
```

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
Fungrim: If there is I can't seem to find it. Will supply it in the java docs instead.

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
Fungrim: Not entirely sure what you mean. They are used in the config class.

logging/README.md
loicmathieu: Usually we check that OT (or open tracing) is present (using build time Capacities) and if so we add the needed code via a build step.

logging/README.md
loicmathieu: OK, understood.

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
loicmathieu: Yes, my bad

logging/deployment/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-core-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-grpc-common-deployment</artifactId>
+        </dependency>
```
loicmathieu, line 22: You should include the `io.quarkiverse.googlecloudservices:quarkus-google-cloud-common-grpc-deployment` dependency instead.

logging/runtime/pom.xml
loicmathieu: It's the other way around ;)
You should remove this dep and add the `io.quarkiverse.googlecloudservices:quarkus-google-cloud-common-grpc-deployment` dependency to the deployment module.

loicmathieu (changes requested):

logging/README.md
Fungrim: Ok, shall we save that for a future improvement?

logging/runtime/pom.xml
Fungrim: Aaaand... Where can I find this `quarkus-google-cloud-common-grpc-deployment` thingy? If I try to add it with version `${project.version}` I get an artifact not found error. It feels like I'm missing something :-D

logging/README.md
loicmathieu: Yes, let's not add things to the already big PR ;)

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
loicmathieu: The way to provide allowed values is via an enum.

logging/deployment/pom.xml
loicmathieu: There is no such deployment dependency, sorry I was measleading. It's OK.

logging/runtime/pom.xml
loicmathieu: I was misleading, you do need quarkus-grpc-common on your runtime module and quarkus-grpc-common-deployment on you deployment module.

The `quarkus-google-cloud-common-grpc` is a runtime only dependency for specific Google Cloud gRPC stuff (native support).

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
Fungrim: Sure, and `LogFormat` is an enum. I was hoping `ConfigItem` had a field for allowed values though.

loicmathieu:
@Fungrim do you still want to contribute this feature ? If will you have time to end this ?
If not may I take your contribution and end it myself ?

loicmathieu:
@Fungrim as I didn't receive any update from you I guess you don't have time to finish it so I'll take it over.
Anyway, thanks for the work you did, very appreciated !

PR #234, opened 2022-01-28
Define Assertj version

gsmet:


loicmathieu:
Thanks, I planned to do it today but you was quicker at it ;)

gsmet:
Yeah I did all of them in a row, it was easier this way.

PR #386, opened 2023-02-15
Operations logging

### LINKED ISSUE #146: Extension for setup GCP logging
Would it be possible to configure Quarkus Logging for GCP logging ?
As quarkus ignore java.util.logging.config.file, it is not possible to configure it with the GCP LoggingHandler like explianed in https://cloud.google.com/logging/docs/setup/java#the_javautillogging_handler
The only workaround I found, is to register and configure the handler manually ...
The console handler produces logs but the level of the log message is not set in the severity of the LogEntry, etc ...

But maybe I missed something ...
Otherwise, a new extension would be nice, as it was suggested in https://stackoverflow.com/questions/67870277/is-there-is-a-way-to-set-up-google-cloud-logging-for-quarkus#comment119986717_67870277 :)

loicmathieu:
Fixes #225 
Fixes #146


loicmathieu:
Thanks again @Fungrim for the initial contribution! Your changes are now merged ;)

PR #414, opened 2023-04-18
Add a missing substitution for Netty InternalLoggerFactory

loicmathieu:


loicmathieu:
This substitution is in fact not needed as we exclude Netty shaded to replace it with Quarkus Netty.

PR #439, opened 2023-05-30
feature(devservices): added pubsub dev service

zZHorizonZz:
Hey there!

Just as I promised, I'm presenting the experimental version of first pubsub devservice, which I already had set up and ready to go. 

This specific devservice is based on the test container implementation of a pubsub emulator, and it's not the only one we can work on using test containers. Here are a few more we can use:

- Bigtable
- Datastore
- Firestore
- Spanner

When it comes to Firebase, though, things may get a little bit more challenging. Unfortunately, test containers don't come with a built-in Firebase implementation, so we'll have to build one from scratch via GenericContainer.

~~This devservice doesn't yet include integration tests.~~

Looking forward to hearing your thoughts on this!

### DIFF
--- a/docs/modules/ROOT/pages/pubsub.adoc
+++ b/docs/modules/ROOT/pages/pubsub.adoc
@@ -203,7 +203,7 @@ If we want to connect to the Dev Service, we need to specify `TransportChannelPr
 
 We can just reuse the code from the previous example and add the `TransportChannelProvider` to the `Subscriber` and `Publisher`. So what do we need to change?
 
-As a first thing we should declare variable which we can then reuse:
+As a first thing, we should declare a variable which we can then reuse:
 
 [source,java]
 ----
--- a/docs/modules/ROOT/pages/pubsub.adoc
+++ b/docs/modules/ROOT/pages/pubsub.adoc
@@ -199,149 +199,18 @@ You can also set the
 
 === Using the Dev Service
 
-If we want to connect to the Dev Service, we need to specify `ChannelProvider` when creating subscription and publisher.
+If we want to connect to the Dev Service, we need to specify `TransportChannelProvider` when creating subscription and publisher.
 
-For this example, we will use the code from the previous example, and we will add a `ChannelProvider` to the `Publisher` and `Subscriber` builders.
+We can just reuse the code from the previous example and add the `TransportChannelProvider` to the `Subscriber` and `Publisher`. So what do we need to change?
+
+As a first thing we should declare variable which we can then reuse:
 
 [source,java]
 ----
-import java.io.IOException;
-import java.util.Optional;
-import java.util.concurrent.TimeUnit;
-import java.util.stream.StreamSupport;
-
-import javax.annotation.PostConstruct;
-import javax.annotation.PreDestroy;
-import javax.ws.rs.GET;
-import javax.ws.rs.Path;
-import javax.ws.rs.Produces;
-import javax.ws.rs.core.MediaType;
-
-import org.eclipse.microprofile.config.inject.ConfigProperty;
-import org.jboss.logging.Logger;
-
-import com.google.api.core.ApiFuture;
-import com.google.api.core.ApiFutureCallback;
-import com.google.api.core.ApiFutures;
-import com.google.api.gax.core.CredentialsProvider;
-import com.google.api.gax.grpc.GrpcTransportChannel;
-import com.google.api.gax.rpc.FixedTransportChannelProvider;
-import com.google.api.gax.rpc.TransportChannelProvider;
-import com.google.cloud.pubsub.v1.MessageReceiver;
-import com.google.cloud.pubsub.v1.Publisher;
-import com.google.cloud.pubsub.v1.Subscriber;
-import com.google.cloud.pubsub.v1.SubscriptionAdminClient;
-import com.google.cloud.pubsub.v1.SubscriptionAdminSettings;
-import com.google.common.util.concurrent.MoreExecutors;
-import com.google.protobuf.ByteString;
-import com.google.pubsub.v1.ProjectName;
-import com.google.pubsub.v1.ProjectSubscriptionName;
-import com.google.pubsub.v1.PubsubMessage;
-import com.google.pubsub.v1.PushConfig;
-import com.google.pubsub.v1.Subscription;
-import com.google.pubsub.v1.Topic;
-import com.google.pubsub.v1.TopicName;
-
-import io.grpc.ManagedChannel;
-import io.grpc.ManagedChannelBuilder;
-
-@Path("/pubsub")
-public class PubSubResource {
-    private static final Logger LOG = Logger.getLogger(PubSubResource.class);
-
-    @ConfigProperty(name = "quarkus.google.cloud.project-id")
-    String projectId;// Inject the projectId property from application.properties
-
-    private TopicName topicName;
-    private Subscriber subscriber;
-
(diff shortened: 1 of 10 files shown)


loicmathieu:
Hi,
Thanks for your contribution, I didn't check on details but this looks great and I'll be happy to accept it.
However, I notice some defaults that raises questions:
- Maybe it should be disabled by default as I don't know if most people would use emulators
- The fixed port bother me a little, I think some day we may want to introduce some sort of global `PubSub` object that allows to more easily create publishers and consumers. Can you set the default to 0 - dynamic port so that when we have this global `PubSub` object it will be automatically setup with the correct dynamic port ? 
- Why are you using by default an image that is not the image used by the testcontainer gcloud module ?

zZHorizonZz:
> Maybe it should be disabled by default as I don't know if most people would use emulators
> The fixed port bother me a little, I think some day we may want to introduce some sort of global PubSub object that allows to more easily create publishers and consumers. Can you set the default to 0 - dynamic port so that when we have this global PubSub object it will be automatically setup with the correct dynamic port ?


I will look into that. 

> Why are you using by default an image that is not the image used by the testcontainer gcloud module ?

While working on this, I was investigating other devservices (how they work, etc.), and noticed they have configurations for images so I added them as well. Regarding the default value, it's an image used within the PubSubEmulatorContainer itself. I simply specified the value that's used in an example of containers found [here](https://www.testcontainers.org/modules/gcloud/)

```java
private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse(
        "gcr.io/google.com/cloudsdktool/google-cloud-cli"
    );
```

zZHorizonZz:
@loicmathieu, should I also add a dedicated documentation page for development services? Or should I incorporate the configuration details for each development service into their respective extension documentation?

loicmathieu:
@zZHorizonZz please add a section on devservices on the respective existing extension page for each service.
I'll add a section later on the main page if it makes sense.

zZHorizonZz:
@loicmathieu, okay, I'll add that. While I'm at it, I've noticed two things about the Quarkus extensions. First, the guide with relevant link configuration is missing in quarkus-extension.yaml. Second, the Firebase Admin extension is named "Google Cloud Firebase Auth," which seems odd to me personally, as it provides more than just authentication. I believe it should be renamed as either "Firebase Admin" or "Google Cloud Firebase Admin".

zZHorizonZz:
@loicmathieu, I've added the devservice documentation. In addition, I've added a guide and configuration to the PubSub quarkus-extension.yml, as this PR is specifically related to PubSub. For other extensions, I'll create new PRs, or these additions can be integrated gradually with the next devservices.

loicmathieu:
@zZHorizonZz thanks a lot for doing this, at first look this seems good, I'll review it in detail next week.

For the missing guide inside the quarkus-extension.yml you can open a PR for adding it for all extensions or I'll do it later.

pubsub/deployment/src/main/java/io/quarkiverse/googlecloudservices/pubsub/deployement/PubSubDevServiceProcessor.java
```diff
@@ -0,0 +1,168 @@
...
+ */
+@BuildSteps(onlyIfNot = IsNormal.class, onlyIf = GlobalDevServicesConfig.Enabled.class)
+public class PubSubDevServiceProcessor {
+
+    private static final Logger LOGGER = Logger.getLogger(PubSubDevServiceProcessor.class.getName());
+
+    // Running dev service instance
+    static volatile DevServicesResultBuildItem.RunningDevService devService;
+    // Configuration for the PubSub Dev service
+    static volatile PubSubDevServiceConfig config;
```
loicmathieu, line 33: Can be private

pubsub/deployment/src/main/java/io/quarkiverse/googlecloudservices/pubsub/deployement/PubSubDevServiceProcessor.java
```diff
@@ -0,0 +1,168 @@
...
+    public DevServicesResultBuildItem startPubSub(DockerStatusBuildItem dockerStatusBuildItem,
+            PubSubBuildTimeConfig pubSubBuildTimeConfig,
+            List<DevServicesSharedNetworkBuildItem> devServicesSharedNetworkBuildItem,
+            Optional<ConsoleInstalledBuildItem> consoleInstalledBuildItem,
+            CuratedApplicationShutdownBuildItem closeBuildItem,
+            LaunchModeBuildItem launchMode,
+            LoggingSetupBuildItem loggingSetupBuildItem,
+            GlobalDevServicesConfig globalDevServicesConfig) {
+        // If dev service is running and config has changed, stop the service
+        if (devService != null && !pubSubBuildTimeConfig.devservice.equals(config)) {
```
loicmathieu, line 45: It didn't seems you set the `config` object anywhere.

pubsub/deployment/src/main/java/io/quarkiverse/googlecloudservices/pubsub/deployement/PubSubDevServiceProcessor.java
```diff
@@ -0,0 +1,168 @@
...
+                config.port.orElse(null));
+
+        // Set container startup timeout if provided
+        timeout.ifPresent(pubSubEmulatorContainer::withStartupTimeout);
+        pubSubEmulatorContainer.start();
+
+        // Return running service item with container details
+        return new DevServicesResultBuildItem.RunningDevService(PubSubBuildSteps.FEATURE,
+                pubSubEmulatorContainer.getContainerId(),
+                pubSubEmulatorContainer::close, "pubsub", pubSubEmulatorContainer.getEmulatorEndpoint());
```
loicmathieu, line 121: The `key` should be a config key, we don't have any for PubSub but it will be better to follow the naming and use something like `quarkus.google.cloud.pubsub.emulator-host`, it could then be used to init the `TransportChannelProvider`

docs/modules/ROOT/pages/pubsub.adoc
```diff
@@ -184,3 +184,69 @@ public class PubSubResource {
...
+
+* `quarkus.google.cloud.pubsub.devservice.enabled` property to `true`
+
+You can also set the
+
+* `quarkus.google.cloud.pubsub.devservice.port` property to change the port on which the emulator will be started (by default there is no port set, so the emulator will use a random port)
+
+=== Using the Dev Service
+
+If we want to connect to the Dev Service, we need to specify `TransportChannelProvider` when creating subscription and publisher.
```
loicmathieu, line 202:
```suggestion
If we want to connect to the Dev Service, we need to specify `TransportChannelProvider` when creating subscriptions and publishers.
```

docs/modules/ROOT/pages/pubsub.adoc
```diff
@@ -184,3 +184,69 @@ public class PubSubResource {
...
+
+We can just reuse the code from the previous example and add the `TransportChannelProvider` to the `Subscriber` and `Publisher`. So what do we need to change?
+
+As a first thing we should declare variable which we can then reuse:
```
loicmathieu, line 206:
```suggestion
As a first thing, we should declare a variable which we can then reuse:
```

loicmathieu (changes requested):

loicmathieu (approved):
LGTM, thanks a lot

PR #455, opened 2023-06-18
Add option to configure GCP logging extension's logging target

### LINKED ISSUE #453: Add option to write to stdout in Google Cloud Logging
Hey,
Google Cloud Logging offers great support for writing JSON logs with traces and stracktraces to GCP via Google Cloud Operations API. But there is a problem if you run the application in the GCP environment as you have duplicated logs: ones from stdout and ones from API. It would be great to add support that we flush structured JSONs to stdout and let Google take care of the ingestion. Something similar is supported in the official Java Google Cloud Logging [library](https://github.com/googleapis/java-logging#alternative-way-to-ingest-logs-in-google-cloud-managed-environments).

I would be happy to implement this myself, but I want to check if this would be merged before I start on implementation.

zanmagerl:
This resolves #453.

I introduced two new configuration options:

- `enableConsoleLogging`: with this user can enable or disable default Quarkus console logging. By default, this is disabled so that we prevent duplicated logs.
- `logTarget`: with this user can configure the target of the logs: stdour, stderr, or Google Cloud API. It follows a similar approach as in the official Google Java Logging library (we even use their enums to avoid having to add additional boilerplate code).

With this user has the option to configure logging targets as they wish. They can output logs to Cloud Logging API and still use Quarkus console logging or they disable console logging in case they do not need them or they want to avoid duplication of logs (if they host their app on Google Cloud Platform).

Testing: using Quarkus getting-started sample on local and on GCP environment.

### DIFF
--- a/logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
+++ b/logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
@@ -3,6 +3,7 @@ package io.quarkiverse.googlecloudservices.logging.runtime;
 import java.util.Map;
 import java.util.Optional;
 
+import com.google.cloud.logging.LoggingHandler.LogTarget;
 import com.google.cloud.logging.Severity;
 import com.google.cloud.logging.Synchronicity;
 
@@ -11,7 +12,7 @@ import io.quarkus.runtime.annotations.ConfigItem;
 import io.quarkus.runtime.annotations.ConfigPhase;
 import io.quarkus.runtime.annotations.ConfigRoot;
 
-@ConfigRoot(name = "google.cloud.logging", phase = ConfigPhase.RUN_TIME)
+@ConfigRoot(name = "google.cloud.logging", phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
 public class LoggingConfiguration {
 
     /**
@@ -25,6 +26,12 @@ public class LoggingConfiguration {
      */
     @ConfigItem(defaultValue = "true")
     public boolean enabled;
+
+    /**
+     * Enable or disable default Quarkus console logging.
+     */
+    @ConfigItem
+    public boolean enableConsoleLogging;
 
     /**
      * Configure base formatting to be either plain text or
@@ -73,6 +80,14 @@ public class LoggingConfiguration {
      */
     @ConfigItem
     public StructuredConfig structured;
+
+    /**
+     * Configures if logs should be written to stdout or stderr instead of using Google Cloud Operations API.
+     * Useful if app is deployed to managed Google Cloud Platform environment with installed logger agent.
+     * Possible values: STDOUT, STDERR and CLOUD_LOGGING.
+     */
+    @ConfigItem(defaultValue = "STDOUT")
+    public LogTarget logTarget;
 
     @ConfigGroup
     public static class StructuredConfig {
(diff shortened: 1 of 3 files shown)


logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
```diff
@@ -74,6 +81,14 @@ public class LoggingConfiguration {
     @ConfigItem
     public StructuredConfig structured;
 
+    /**
+     * Configures if logs should be written to stdout or stderr instead of using Google Cloud Operations API.
+     * Useful if app is deployed to managed Google Cloud Platform environment with installed logger agent.
+     * Possible values: STDOUT, STDERR and CLOUD_LOGGING.
+     */
+    @ConfigItem(defaultValue = "STDOUT")
+    public LogTarget logTarget;
```
loicmathieu, line 90: Default should be CLOUD_LOGGING as it was the default before this PR.

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
```diff
@@ -26,6 +27,12 @@ public class LoggingConfiguration {
     @ConfigItem(defaultValue = "true")
     public boolean enabled;
 
+    /**
+     * Enable or disable default Quarkus console logging.
+     */
+    @ConfigItem
+    public boolean enableConsoleLogging;
```
loicmathieu, line 34: This config property is not used anywhere

loicmathieu (changes requested):

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
zanmagerl: of course, will change :)

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
zanmagerl: it is used, inside `LoggingBuildSteps`, where we configure if we want to disable Quarkus console logging

loicmathieu: Yes but it is not used in runtime code so it seems useless, the logTarget is enough IMHO

loicmathieu (changes requested):

logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingConfiguration.java
zanmagerl: It is not useless because it disables console logging with which we avoid duplicated logs. This GCP extension is currently registering another log handler, which causes us to have 2 logging handlers and 2 log outputs. With this option, we have the option to disable this by default so that user does not have to disable them manually (but they still have the option to keep console logging if they want).

loicmathieu: Just understood what you did!

If someone want to disable console login, he can use the `quarkus.log.console.enable` property, I don't see the point of providing another config property that will set this one for him.

loicmathieu (approved):
LGTM, thanks

PR #465, opened 2023-07-14
Use `graal-sdk` instead of `svm`

gastaldi:
- Caused by https://github.com/quarkusio/quarkus/pull/34145

### DIFF
--- a/common/grpc/pom.xml
+++ b/common/grpc/pom.xml
@@ -55,8 +55,8 @@
             to be able to support 22.2 and 22.3.
         -->
         <dependency>
-            <groupId>org.graalvm.nativeimage</groupId>
-            <artifactId>svm</artifactId>
+            <groupId>org.graalvm.sdk</groupId>
+            <artifactId>graal-sdk</artifactId>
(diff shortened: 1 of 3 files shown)


common/grpc/pom.xml
```diff
@@ -55,8 +55,8 @@
             to be able to support 22.2 and 22.3.
         -->
         <dependency>
-            <groupId>org.graalvm.nativeimage</groupId>
-            <artifactId>svm</artifactId>
+            <groupId>org.graalvm.sdk</groupId>
+            <artifactId>graal-sdk</artifactId>
```
loicmathieu, line 59: The GraalVM SDK is already included just under.
Can you remove the duplicated dependency please ?

loicmathieu (changes requested):

common/grpc/pom.xml
gastaldi: Doh, I forgot to check that in my script, thanks for catching it!

loicmathieu (approved):
LGTM, thanks

PR #476, opened 2023-08-07
Replace deprecated Config APIs

radcortez:


common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpBootstrapConfiguration.java
```diff
@@ -2,44 +2,43 @@
...
+    boolean enableMetadataServer();
 
     /**
      * Google Cloud project ID.
      * It defaults to `ServiceOptions.getDefaultProjectId()` if `google.cloud.enable-metadata-server` is set to true (which is
      * the default),
      * so to the project ID corresponding to the default credentials if the default credentials are set, otherwise null.
      */
-    @ConfigItem(defaultValueDocumentation = "ServiceOptions.getDefaultProjectId()")
-    public Optional<String> projectId;
```
loicmathieu, line 30: With ConfigMapping there is no way to document the default value anymore?

common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpDefaultsConfigSourceFactory.java
```diff
@@ -0,0 +1,35 @@
...
+        ConfigValue enableMetadataServer = context.getValue("quarkus.google.cloud.enable-metadata-server");
+        if (enableMetadataServer.getValue() != null) {
+            if (Converters.getImplicitConverter(Boolean.class).convert(enableMetadataServer.getValue())) {
+                String defaultProjectId = ServiceOptions.getDefaultProjectId();
+                if (defaultProjectId != null) {
+                    return singletonList(
+                            new PropertiesConfigSource(Map.of("quarkus.google.cloud.project-id", defaultProjectId),
+                                    "GcpDefaultsConfigSource",
+                                    -Integer.MAX_VALUE));
+                }
```
loicmathieu, line 30: The default projectId should only be used when there is no supplied projectId.
Can you confirm that this will be the case for this piece of code?
I assume `-Integer.MAX_VALUE` is the config source order so it will be take last?

radcortez: Yes, the idea is that it will add the `projectId` in a lower ordinal source, so if the `projectId` is set in any of the traditional source, it will take precedence over this one.

loicmathieu (approved):
LGTM, just tested it locally and it works like a charm, thanks.

loicmathieu:
@all-contributors please add @radcortez for code

PR #495, opened 2023-09-08
Fix a11y-base version to avoid dependency convergence issue

loicmathieu:


gsmet:
You shouldn't need to do this at your level I think.

@phillip-kruger told me it was fixed in recent version and for latest 3.3, I think I ended up adding the dependency in the BOM.

With which version do you have this issue?

loicmathieu:
There were ecosystem CI failures.

There is other issues with 3.4 that I need to sort out, I'll remove the dependency convergence when doing the 3.4 migration.

gsmet:
I wouldn't expect problems with `main` as @phillip-kruger told me this was fixed. @phillip-kruger can you confirm?

phillip-kruger:
Yes main should be fixed

loicmathieu:
I removed it and upgrade to Quarkus 3.4.0.CR1, everything works.

PR #499, opened 2023-09-11
Do not use CDI beans to read the SM

loicmathieu:


radcortez:
Sure. Sorry for the noise.

loicmathieu:
Not a problem, this one is a complex beast as it may be used in very early stage of the application and as I'm lazy and don't want to configure a specific CI with access to Google Cloud I find it easier to disable it on CI.

PR #487, opened 2023-08-23
Prevent the usage of OpenTelemetry in GcpDefaultsConfigSourceFactory

geoand:
This is needed because GcpDefaultsConfigSourceFactory is called very early in the application startup
and can thus mess up the OpenTelemetry configuration performed later by Quarkus

Fixes: https://github.com/quarkusio/quarkus/issues/35500

P.S. This pattern is also used in Quarkus code [here](https://github.com/quarkusio/quarkus/blob/1eb8b2d80673abe76b9cced7a1cf491872d22748/extensions/container-image/container-image-jib/deployment/src/main/java/io/quarkus/container/image/jib/deployment/JibProcessor.java#L246-L267).

PR #502, opened 2023-09-14
Do not use Google's HTTP client to get the default project ID

snazy:
As described in [this issue](https://github.com/quarkusio/quarkus/issues/35500), Quarkus w/ the Google clooud services and OpenCensus shim does not startup due to a static initialization issue.

`com.google.cloud.ServiceOptions` is used by the extension to get the Google cloud default-project-ID, which uses Google's HTTP client, which uses OpenCensus, which triggers an initialization of OpenTelemetry, which conflicts with the OTel init from Quarkus.

This change updates `GcpDefaultsConfigSourceFactory` to use Java's HTTP client.

See also #487, the alternatives mentioned in [this comment](https://github.com/quarkiverse/quarkus-google-cloud-services/pull/487#issuecomment-1700701468) to implement a `ConfigSource` and in [this comment](https://github.com/quarkiverse/quarkus-google-cloud-services/pull/487#issuecomment-1700784281) to set `otel.java.global-autoconfigure.enabled=false)` end in the same OpenCensus/OpenTracing init race.

Fixes https://github.com/quarkusio/quarkus/issues/35500


snazy:
Manually verified that Quarkus starts w/ the OpenCensus-shim.

geoand (approved):
Nice!

snazy:
Also manually verified (in `jshell` in a GCP VM) that the `getAppEngineProjectIdFromMetadataServer()` function yields the default project ID

common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpDefaultsConfigSourceFactory.java
```diff
@@ -17,40 +29,138 @@
...
+            String projectIdURL = "/computeMetadata/v1/project/project-id";
+
+            try {
+                URI uri = new URI(metadata + projectIdURL);
+                HttpClient client = HttpClient.newBuilder().connectTimeout(Duration.ofMillis(500)).build();
+                HttpRequest request = HttpRequest.newBuilder().timeout(Duration.ofMillis(500)).GET().uri(uri)
+                        .header("Metadata-Flavor", "Google").build();
+                HttpResponse.BodyHandler<String> bodyHandler = HttpResponse.BodyHandlers.ofString();
+                HttpResponse<String> response = client.send(request, bodyHandler);
+                return headerContainsMetadataFlavor(response) ? response.body() : "";
```
loicmathieu, line 149: Shouldn't it respond null and not empty if not metadata (as the remaining if chain check for nullity)

common/runtime/src/test/java/io/quarkiverse/googlecloudservices/common/GcpDefaultsConfigSourceFactoryTest.java
```diff
@@ -0,0 +1,59 @@
...
+        Mockito.when(context.getValue("quarkus.google.cloud.enable-metadata-server"))
+                .thenReturn(ConfigValue.builder().withValue("false").build());
+
+        Iterable<ConfigSource> configSources = new GcpDefaultsConfigSourceFactory(() -> "test-project-id")
+                .getConfigSources(context);
+        assertThat(configSources).isEmpty();
+    }
+
+    @Test
+    void staticOpenCensusOpenTelemetryInit() {
```
loicmathieu, line 41: What is the purpose of this test? Does it crash without the change (so it's a non-regression test?)

pom.xml
```diff
@@ -21,6 +21,7 @@
     <compiler-plugin.version>3.8.1</compiler-plugin.version>
     <enforcer-plugin.version>3.4.1</enforcer-plugin.version>
     <assertj.version>3.24.2</assertj.version>
+    <opentelemetry-alpha.version>1.28.0-alpha</opentelemetry-alpha.version>
```
loicmathieu, line 24:
```suggestion
    <opentelemetry.version>1.28.0-alpha</opentelemetry.version>
```

And use the same under

loicmathieu (changes requested):

common/runtime/src/main/java/io/quarkiverse/googlecloudservices/common/GcpDefaultsConfigSourceFactory.java
snazy: Ah! Good catch!

common/runtime/src/test/java/io/quarkiverse/googlecloudservices/common/GcpDefaultsConfigSourceFactoryTest.java
snazy: Yes, fails without the prod code change. Added a comment.

common/runtime/pom.xml
```diff
@@ -97,6 +119,15 @@
...
             </plugin>
+<!--                    <plugin>-->
+<!--                        <artifactId>maven-surefire-plugin</artifactId>-->
+<!--                        <configuration>-->
+<!--                            <systemProperties>-->
+<!--                                <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>-->
+<!--                                <gcloud.test>true</gcloud.test>-->
+<!--                            </systemProperties>-->
+<!--                        </configuration>-->
+<!--                    </plugin>-->
```
loicmathieu, line 130: Why do you comment this? If it's not needed you can remove it but as far as I remember this is needed to have logs property formatted on tests

snazy: Left over (sorry)

I initially added it, because I was thinking of some `@QuarkusTest`, but then realized that it's good enough to check the static-init race.

loicmathieu (approved):
LGTM, I'll merge it tomorrow for the next release based on Quarkus 3.4

snazy:
@loicmathieu I personally don't mind (can wait for 3.4, if the 3.4 platform-bom references it), but could be useful for other users to have this also in 3.3.x.

loicmathieu:
@all-contributors please add @snazy for code

loicmathieu:
2.5.0 is released with the fix.
If there is a strong need to backport it ot 2.4.0 (which is based on Quarkus 3.3) I can do it.

PR #504, opened 2023-09-15
feature(devservices): added bigtable dev service

zZHorizonZz:
This PR adds devservice for bigtable. Similar to firstore or pubsub emulators.

### DIFF
--- a/bigtable/deployment/pom.xml
+++ b/bigtable/deployment/pom.xml
@@ -27,6 +27,10 @@
             <groupId>io.quarkiverse.googlecloudservices</groupId>
             <artifactId>quarkus-google-cloud-bigtable</artifactId>
         </dependency>
+        <dependency>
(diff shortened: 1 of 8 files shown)


zZHorizonZz:
I'm now wondering Is there a reason why there isn't a Bigtable/admin client producer?

loicmathieu (approved):
LGTM, thanks a lot

loicmathieu:
> I'm now wondering, Is there a reason why there isn't a Bigtable/admin client producer?

At the time being, it was because there is not a single client but multiple ones.
I think adding a producer for the client would be a good addition.

zZHorizonZz:
> At the time being, it was because there is not a single client but multiple ones. I think adding a producer for the client would be a good addition.

OK then I will try to look into it and with it, I will also add the documentation for this devservice.

PR #508, opened 2023-09-20
feature(bigtable): added bigtable client creation

zZHorizonZz:
This PR introduces an experimental version of the Bigtable client creation. 

**Summary:**

- Introduced a BigtableClient annotation that qualifies an injected Bigtable client.
- The BigtableClientProcessor class is responsible for registering beans, discovering injected beans, generating bigtable client producers, and transforming injection points. This ensures the right Bigtable client is used based on the configuration and qualifiers.
- The main class BigtableClients handles the creation of the Bigtable client.
- The system for bean creation is based on grpc client bean creation.

Also configuration will be made so that you can configure it like this:

```
quarkus.google.cloud.bigtable.clients.test-client.instance-id=test-instance
```

or default client:

```
quarkus.google.cloud.bigtable.clients.instance-id=test-instance
```

The introduced Bigtable client creation is experimental and may be subject to changes based on further testing.

### DIFF
--- a/bigtable/api/pom.xml
+++ b/bigtable/api/pom.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>quarkus-google-cloud-bigtable-parent</artifactId>
+        <groupId>io.quarkiverse.googlecloudservices</groupId>
+        <version>2.5.0-SNAPSHOT</version>
+    </parent>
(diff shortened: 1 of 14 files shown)


zZHorizonZz:
@loicmathieu When you have a bit of time, could you take a look at this? If the system this PR introduces is okay, can I potentially move forward or should I rework it?

loicmathieu (commented):
Hi,
Thanks for the contribution.
Is there a strong need for multiple client support? It complexifies a lot the PR where a simple approach with a CDI producer would have worked if only one instance is published.

zZHorizonZz:
> Is there a strong need for multiple client support? It complexifies a lot the PR where a simple approach with a CDI producer would have worked if only one instance is published.

So, yeah, that's why I wanted to wait first since you mentioned in https://github.com/quarkiverse/quarkus-google-cloud-services/pull/504#issuecomment-1723174438 something about multiple clients, etc. Also, I'm not sure if it's normal to use multiple Bigtable instances or just one in production. In any case, I will close this PR and create a new one with just a simple producer. 👍

loicmathieu:
Yes, thanks, this would be easier to maintant!

PR #572, opened 2024-02-07
Include docs module as a Maven module

gastaldi:
- This is checked in the release workflow


docs/modules/ROOT/pages/includes/attributes.adoc
```diff
@@ -0,0 +1,3 @@
+:project-version: 2.7.0
```
loicmathieu, line 1: Should this version be updated each time I do a release?

loicmathieu (approved):
Thanks, I notice that doc needs a special treatment now as my last attempt to do a release fail.

docs/modules/ROOT/pages/includes/attributes.adoc
gastaldi: This is updated automatically from the templates/ directory, so you don't need to worry about it

PR #602, opened 2024-03-26
EmulatorCredentials when running on dev services

jfbenckhuijsen:
Implementation to use EmulatorCredentials for dev services (unless access-token-enabled = true)

### DIFF
--- a/firestore/runtime/src/main/java/io/quarkiverse/googlecloudservices/firestore/runtime/FirestoreProducer.java
+++ b/firestore/runtime/src/main/java/io/quarkiverse/googlecloudservices/firestore/runtime/FirestoreProducer.java
@@ -4,6 +4,7 @@ import java.io.IOException;
 
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.inject.Default;
+import jakarta.enterprise.inject.Instance;
 import jakarta.enterprise.inject.Produces;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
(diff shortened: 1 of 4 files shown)


firestore/runtime/src/main/java/io/quarkiverse/googlecloudservices/firestore/runtime/FirestoreProducer.java
```diff
@@ -68,4 +77,16 @@ private Duration convertDuration(java.time.Duration duration) {
...
+    /**
+     * Determine if we need to use emulator credentials. Emulator credentials are used in case the host-override is set
+     * and we don't have accessTokens enabled.
+     *
+     * @return whether to use the emulator credentials
+     */
+    private boolean useEmulatorCredentials() {
+        return !this.gcpBootstrapConfiguration.accessTokenEnabled()
+                && this.firestoreConfiguration.hostOverride.isPresent()
+                && this.firestoreConfiguration.hostOverride.get().contains("localhost");
```
loicmathieu, line 89: Why also checking on localhost?
If an emulator is used for ex on a remote Docker engine, or if the user uses multiple host on its machine it can be on something else than localhost.

jfbenckhuijsen: Ref: com.google.cloud.firestore.spi.v1.GrpcFirestoreRpc:90

loicmathieu: OK, if it is the same inside the Google Cloud library so let's keep it this way, we can always relax it later if needed.

loicmathieu (approved):
LGTM, thanks

loicmathieu:
@all-contributors, please add [jfbenckhuijsen](https://github.com/jfbenckhuijsen) for code

Issue #674, opened 2024-09-19
GCP logging does not include span and sampled fields in structured logging


wabrit:
I'm using version 2.11.0 of the quarkus-google-cloud-logging quarkiverse extension so that I can include tracing information when logging to GCP.

Following the instructions [here](https://docs.quarkiverse.io/quarkus-google-cloud-services/main/logging.html#_tracing) I've implemented a TraceInfoExtractor that returns the current trace and span ids.

However the resulting payload, while it includes the trace in the format GCP expects, does not include the `logging.googleapis.com/spanId` value, and `logging.googleapis.com/trace_sampled` is left defaulted to false.

```json
{
    "severity":"WARNING",
    "time":"2024-09-18T10:31:19.270162900Z",
    "logging.googleapis.com/labels":{},
    "logging.googleapis.com/trace":"projects/myProject/traces/2e6d1e217fe2a1c2c331424649918107",
    "logging.googleapis.com/trace_sampled":false,
       ...
}

```

Without these additional fields, it is not possible to navigate between traces and logs in the GCP console.

As far as I can see the issue is in the `LoggingHandler#transform` method from this extension, which omits to call the builder methods to set these (see java comments):

```java
private LogEntry transform(ExtLogRecord record, TraceInfo trace) {
        Payload<?> payload = internalHandler.transform(record, trace);
        if (payload != null) {
            com.google.cloud.logging.LogEntry.Builder builder = LogEntry.newBuilder(payload)
                    .setSeverity(LevelTransformer.toSeverity(record.getLevel()))
                    .setTimestamp(record.getInstant());

           /*
            * Ideally should be setting labels for logging here too i.e.
            *
            * builder.addLabel("levelName", LevelTransformer.toSeverity(record.getLevel()).toString())
            * builder.addLabel("levelValue", String.valueOf(level.intValue()))
            * builder.addLabel("loggerName", extLogRecord.getSourceClassName())
            */ 

            if (this.config.gcpTracing().enabled() && trace != null && !Strings.isNullOrEmpty(trace.getTraceId())) {
                builder = builder.setTrace(composeTraceString(trace.getTraceId()));

               /* Should also be setting these for integrated log/tracing support:
                 *  builder.setSpan(trace.getSpanId())
                 *  builder.setTraceSampled(true) 
                 */
            }
            return builder.build();
        } else {
            return null;
        }
    }
```

nhatlaso1:
install this
[...]

PR #677, opened 2024-09-20
Issue 674 gcp logging spaninfo

wabrit:
These changes address [issue 674](https://github.com/quarkiverse/quarkus-google-cloud-services/issues/674) by ensuring that the span id and sampled properties are set on the `LogEntry.builder` (without these settings it is not possible to exploit the feature of GCP console where a log entry can be used to launch inspection of the trace corresponding to that log (and vice versa).

These changes have been tested in a native-build Quarkus microservice (running Quarkus 3.11.3).

### DIFF
--- a/logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingHandler.java
+++ b/logging/runtime/src/main/java/io/quarkiverse/googlecloudservices/logging/runtime/LoggingHandler.java
@@ -2,6 +2,7 @@ package io.quarkiverse.googlecloudservices.logging.runtime;
 
 import java.util.Collections;
 import java.util.logging.ErrorManager;
+import java.util.logging.Level;
 
 import org.jboss.logmanager.ExtHandler;
 import org.jboss.logmanager.ExtLogRecord;
@@ -11,6 +12,7 @@ import com.google.cloud.logging.LogEntry;
 import com.google.cloud.logging.Logging;
 import com.google.cloud.logging.Logging.WriteOption;
 import com.google.cloud.logging.Payload;
+import com.google.cloud.logging.Severity;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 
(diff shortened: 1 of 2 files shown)


loicmathieu:
I'm not convinced about the labels, do they bring any added value? I understand that logback do this but for which purpose?

wabrit:
There was one general and one more specific reason I added the labels:

- In a mixed economy of Spring Boot / Quarkus services running in GCP, having consistent GCP stuctured log layout makes inspecting cross-service logs (e.g. in the context of tracing) a little simpler. SB services typically use logback for their GCP logging, so this seemed a way to ensure some degree of consistency.
- The `loggerName` label is probably the most useful of those 3. However when using JSON payload format, the same info and more is admittedly to be found in the `jsonPayload.log.origin.class` field, so the argument for this being repeated here is consistency at best in that when looking at cross-service logs you don't have to know that different components put this info in different places.

In short, I appreciate that there isn't a compelling technical reason for these labels, just a minor-ish usage one. If you'd prefer me to remove them, that's fine.

loicmathieu:
`loggerName` seems legit, the other may be seen as too much duplication but maybe your rational makes sense.

Pinging @Fungrim as he was the original contributor of Google Cloud Logging extension, what do you think about the proposed default labels, does it makes sense?

wabrit:
if it helps I can split out the labels work into a separate PR, and leave this PR to focus on the tracing fix, as the latter is I think of far greater importance.

Fungrim:
Thanks for the ping @loicmathieu, and the span is a nice catch @wabrit.

Regarding the labels and don't have any strong personal opinions. However, given the lack of standardization around log records and formats, I generally dislike hard coded values. If I did this change myself I would probably add a `LogRecordLabelExtractor` interface (much like the trace info extraction) that the user can bind via CDI to add whatever labels they want to the GCP record and/or expose these default values via a configuration option. 

But again, these are not _strong_  so I'm fine with the proposed changes as well.

wabrit:
The `LogRecordLabelExtractor` idea sounds much better than what I've done - thanks @Fungrim!

I'm happy to go either of the following ways:

1. Modify this PR to include `LogRecordLabelExtractor` support
2. Remove label code from this PR and submit a different PR with `LogRecordLabelExtractor` support

@loicmathieu let me know what you would prefer (FWIW my preference would be option (2) as I'd really like to get the trace fix in).

loicmathieu:
@wabrit the best would be to do the 2. : remove the label change from this PR and provide a different PR.

wabrit:
@loicmathieu labels stuff now removed from PR

loicmathieu (approved):
LGTM, thanks a lot

PR #685, opened 2024-10-02
provide an extension mechanism for adding custom labels to gcp logs

wabrit:
This introduces an extension interface `LogRecordLabelExtractor` which can be implemented to allow custom labels to be added to each log entry when logging to gcp.

### DIFF
--- a/docs/modules/ROOT/pages/logging.adoc
+++ b/docs/modules/ROOT/pages/logging.adoc
@@ -143,7 +143,7 @@ By default, this will also be set on the Google Cloud Operations log entry as a 
 
 === Custom Labels
 
-In order to include additional labels in the log, you must bind a `LogRecordLabelSupplier` to the CDI context, e.g.:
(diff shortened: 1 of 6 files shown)


loicmathieu (changes requested):
Hi, thanks for your contribution.
Can you call LogRecordLabelSupplier LogRecordLabelExtractor to be consistent withe the TraceInfoExtractor?

wabrit:
It didn't seem to me that this new one should be an "extractor", in that its function is to _contribute_ additional labels, as opposed to the trace interface which is _extracting_ trace info from the current context.

But it's a minor point and if you prefer similar naming that's fine.

loicmathieu:
But label will most of the cases extracted from the log record not hardcoded (I think we have config props for that) right? So extracted seems to be better

wabrit:
No probs - will change the name as requested

wabrit:
interface now renamed

loicmathieu (approved):
Thanks, LGTM

PR #705, opened 2024-11-01
Use NoCredentials in case the emulator host is being connected to

jfbenckhuijsen:
Dynamically use NoCredentials in case the emulator is used


loicmathieu (approved):
LGTM, thanks

PR #707, opened 2024-11-01
Feature/firebase emulator config switch

jfbenckhuijsen:


.github/workflows/build.yml
loicmathieu, line 1: Can you revert these changes?

loicmathieu (commented):
You create a new extension, why can't the emulator be part of the existing firebase-admin extension? This would simplify a lot this PR.

jfbenckhuijsen:
> You create a new extension, why can't the emulator be part of the existing firebase-admin extension? This would simplify a lot this PR.

Main reason is that it's not just a dev service for firebase-admin. The firebase emulator also support stuff like Pubsub/functions etc, and there are use-cases when it makes sense to combine them based on the platform (the emulator takes care of some integration under the hood which the individual emulators don't).

.github/workflows/build.yml
jfbenckhuijsen: I can, but then I would need to hardcode the user (id) used by GitHub into the code. If you're fine with that I'll revert it.

Main reason for this is that the firebase emulator allows to export and import data via the file system (which is quite convenient to prepare test-data). The only issue is, because the emulator runs as a docker process, you run into filesystem ownership issues. By default, the docker engine on a CI setup does not translate between the UID used inside the docker process and the UID of the user running the container. This means that the files written on the host system will end up with UID/GID as defined by the docker image, meaning that the host process (i.e. the tests) can't access (or even delete) these files.

loicmathieu: I wonder if it cannot have security implication.

@geoand as guardian of the Quarkiverse, WDYT of exporting the user and group via env var?

firebase/deployment/pom.xml
```diff
@@ -0,0 +1,138 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.googlecloudservices</groupId>
+            <artifactId>quarkus-google-cloud-common-grpc</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-firestore</artifactId>
```
loicmathieu, line 64: Why do you need firestore and pubsub in firebase admin?
It's annoying as it brings a lot of dependencies inside it and duplicate dependency managements, if really needed this needs to be Quarkus deps not directly the Google Cloud deps

.github/workflows/build.yml
geoand: You are looking for @gastaldi actually 😎

firebase/deployment/pom.xml
loicmathieu, line 1: It's only  a devservices so it should be named `firebase-devservices`.

firebase/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseDevServiceConfig.java
```diff
@@ -0,0 +1,282 @@
...
+            boolean preferFirebaseDevServices();
+
+            /**
+             * Sets the Docker image name for the Google Cloud SDK.
+             * This image is used to emulate the Pub/Sub service in the development environment.
+             * The default value is 'gcr.io/google.com/cloudsdktool/google-cloud-cli'.
+             * <p>
+             * See also the documentation on Custom Docker images for more info about this image.
+             */
+            @WithDefault("node:23-alpine")
```
loicmathieu, line 97: The configured default didn't match the documentation default

firebase/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseDevServiceConfig.java
```diff
@@ -0,0 +1,282 @@
...
+            String imageName();
+
+            /**
+             * Id of the docker user to run the firebase executable. This is needed in environments where Docker
+             * does not perform a mapping to the user running Docker. In a Docker Desktop setup, Docker
+             * automatically performs this mapping and the data written by the emulator can be read by the user
+             * running the build. This is not the case in a regular (non-Desktop) setup, this is not the case
+             * and you may need to set the user id and {@link #dockerGroup()}. This option is often needed in CI
+             * environments.
+             */
```
loicmathieu, line 107:
```suggestion
            /**
             * Id of the docker user to run the firebase executable. This is needed in environments where Docker
             * does not perform a mapping to the user running Docker. In a Docker Desktop setup, Docker
             * automatically performs this mapping and the data written by the emulator can be read by the user
             * running the build. This is not the case in a regular (non-Desktop) setup, 
             * so you may need to set the user id and {@link #dockerGroup()}. This option is often needed in CI
             * environments.
             */
```

firebase/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseDevServiceConfig.java
```diff
@@ -0,0 +1,282 @@
...
+ * Here is an example of how to configure these properties:
+ * <p>
+ *
+ * <pre>
+ * quarkus.google.cloud.firebase.devservice.enabled = true
+ * quarkus.google.cloud.pubsub.devservice.image-name = gcr.io/google.com/cloudsdktool/google-cloud-cli # optional
+ * quarkus.google.cloud.pubsub.devservice.emulatorPort = 8085 # optional
+ * </pre>
+ */
+@ConfigMapping(prefix = "quarkus.google.cloud")
```
loicmathieu, line 23:
```suggestion
@ConfigMapping(prefix = "quarkus.google.cloud.firebase-dev-services")
```

loicmathieu (changes requested):

.github/workflows/build.yml
loicmathieu: Ah yes, sorry, bad github autocompletion selection ;)

gastaldi: I don't think there is a security implication in this, plus I wouldn't know what to use as the env vars values if they were created

loicmathieu: OK thanks, let's keep  it then.

firebase/runtime/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>quarkus-google-cloud-firebase</artifactId>
+    <name>Quarkus - Google Cloud Services - Firebase - Runtime</name>
+    <description>Use Google Cloud Firebase</description>
+
+    <properties>
+        <firebase-admin-sdk.version>9.4.0</firebase-admin-sdk.version>
+        <microprofile-jwt-auth-api.version>2.1</microprofile-jwt-auth-api.version>
+    </properties>
```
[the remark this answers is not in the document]
jfbenckhuijsen: Ah lol, copy/paste error. They were actually not used at all. Removed them.

firebase/deployment/pom.xml
jfbenckhuijsen: They are there for testing purposes only (they are all test scope, except pubsub, which I missed, this is fixed now). 

Main reason why I included these and not the quarkus versions is this might introduce a circular dependency. Also. the focus of the testing is to validate just the Firebase TestContainer. In the future, we might even split that off into a separate library, as that part has no dependency on quarkus.

firebase/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseDevServiceConfig.java
jfbenckhuijsen: Good catch! Fixed

firebase/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseDevServiceConfig.java
loicmathieu: You can read a config from another extension in an extension.

But we may need an SPI here if we need to access something from this extension from other extensions.
An SPI module is module that expose an autonomous module (with no dependencies) to other extensions for ex to access shared configuration or information. Usually in the form of a build item.

As this PR is already big, we can do this later, but the config root must change

jfbenckhuijsen: Ok clear, can work on that as a next PR. Working on a follow up anyway, so should be able to do that fairly quickly. Can we keep the config root for this PR? If we change it, the setup won't work and it'll result in conflicts with the Firestore and Pubsub dev services.

jfbenckhuijsen: @loicmathieu any update? In the mean time I'm finishing up work on a next PR to be able to include the firebase.json (and other files used by a regular emulator) inside the docker container (easing development). Did some refactoring too which seperates out the TestContainers stuff from the Quarkus dev service/integration stuff. (See https://github.com/jfbenckhuijsen/quarkus-google-cloud-services/pull/1)

Next up will be a PR to get that SPI Module running. Just for my understanding, that's just an extension module (set of -deployment and -runtime) without dependencies to other modules which is used by others?

I suggest I'll create the SPI stuff as a PR onto this PR, so we can split up reviewing into separate PRs just like the one I mentioned in this comment. Once approved, I'll merge them into this PR, and we can merge this one into the main branch?

loicmathieu: > Can we keep the config root for this PR?

No, it really need to use a separate root config, but as I said, you can access any config option from any dependency so this should not be an issue.

loicmathieu:
@jfbenckhuijsen I'll try to review this PR soon so we can proceed with it.
Sorry for the delay, I was very busy lately.

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseDevServiceConfig.java
```diff
@@ -0,0 +1,282 @@
...
+ *
+ * <pre>
+ * quarkus.google.cloud.firebase.devservice.enabled = true
+ * quarkus.google.cloud.pubsub.devservice.image-name = gcr.io/google.com/cloudsdktool/google-cloud-cli # optional
+ * quarkus.google.cloud.pubsub.devservice.emulatorPort = 8085 # optional
+ * </pre>
+ */
+@ConfigMapping(prefix = "quarkus.google.cloud")
+@ConfigRoot
+public interface FirebaseDevServiceConfig {
```
loicmathieu, line 25: Config root must be `quarkus.google.cloud.devservices` so these configs are scoped to the devservices.

The projectId config already exist and should be reused from the common config.

Another solution would be to add the devservices config to each separate module, and if the config is used but the firebase devservices module not included, throw an error at deployment time saying that the firebase devservices module must be included to be able to use pubsub devservcies.

Here there is clearly a dependency between each module that is not correct.
We should take as inspiration how it is done in the AWS service extension pack, however this is a bit different as all devservices are based on localstack, but here some devservices are autonomous and others are based on firebase. https://docs.quarkiverse.io/quarkus-amazon-services/dev/dev-services.html

docs/modules/ROOT/pages/firebase.adoc
```diff
@@ -0,0 +1,98 @@
...
+* Each of the emulators must be exposed on `0.0.0.0` as host as described https://firebase.google.com/docs/emulator-suite/use_hosting#emulators-no-local-host[here]. If this is not done, the Emulators will not be reachable from the Docker host.
+* Emulators need to be configured to use the default ports. Customizing the ports on which they run is currently not supported (this might change in a future version).
+
+== Interaction with other other extensions
+
+The following extensions support a DevService which conflicts with the DevService exposed by the Firebase Emulators.
+
+* Firestore
+* PubSub
+* TODO: Verify Storage
```
loicmathieu, line 92: Better to verify it as TODO in doc is not a good idea.

docs/modules/ROOT/pages/firebase.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+When including this module, these DevServices will automatically be disabled, as the Firebase emulator should feature wise be on-par or more extensive than the individual emulators.
```
loicmathieu, line 94:
```suggestion
When including this module, these  Dev Services will automatically be disabled, as the Firebase emulator should feature wise be on-par or more extensive than the individual emulators.
```

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseEmulatorContainer.java
```diff
@@ -0,0 +1,519 @@
...
+            // TODO: Validate if a custom firebase.json is defined, that the hosts are defined as 0.0.0.0
+        }
+
+        private void configureBaseImage() {
+            dockerBuilder.from(firebaseConfig.dockerConfig().imageName());
+        }
+
+        private void initialSetup() {
+            dockerBuilder
+                    .run("apk --no-cache add openjdk11-jre bash curl openssl gettext nano nginx sudo && " +
```
loicmathieu, line 273: Can it run with OpenJDK 17?

jfbenckhuijsen:
> @jfbenckhuijsen I'll try to review this PR soon so we can proceed with it. Sorry for the delay, I was very busy lately.

tnx, you had me worried ;)

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseEmulatorContainer.java
jfbenckhuijsen: Should be able to, I'll bump it.

docs/modules/ROOT/pages/firebase.adoc
jfbenckhuijsen: Will look into that!

Generic question: in the mean time, I've been doing some work to create a separate module for the firebase-testcontainers stuff (i.e. move the definition of the Container to a separate module). It lives currently at https://github.com/alfa1-group/testcontainers-firebase 

I could change the current PR to use that module as dependency (reducing the size of the PR greatly, the majority of the work is in that container actually). 

Any pref on that?

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseDevServiceConfig.java
jfbenckhuijsen: Ok, using a `q.g.c.devservices` config root seems most sensible to me to avoid all the extra dependencies you would get for including the firebase module. Let me work on that.

As for using the project common config interface, please check the issue raised at https://github.com/quarkiverse/quarkus-google-cloud-services/pull/707#discussion_r1882549543

docs/modules/ROOT/pages/firebase.adoc
loicmathieu: This is not an east to answer question ;) 

Usually, using an existing dependency is geat, but the bar to include a dependency in Quarkus is high so we want each dependency to have clear maintenance and stewardship, regular updates, security quality, ...

So a brand new repo from a not well known source didn't meet the bar for inclusion in Quarkus as a dependency, I hop you understand that I'm not judging the quality of the repo, the maintainer or the company behind it.

If you want to reuse the container, it may be possible to release it as a separate module with no dependencies so it could be reused outside this library.

jfbenckhuijsen: All good and fully understandable. What I'll do I copy the code from the repo to the PR and keep them in sync for now. 

Once the other repo gains enough credibility we can always choose to switch to that (or not and just keep the copied code as is, either is fine).

The other repo has already proven valuable anyhow as it allowed me to start using the code in a real life project and iron out some missing features (and a few small bugs), as well as improve on the api of the module itself. 

I'll finish up on a few of those changes, then copy the altered code back so we can finish this PR.

firebase/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseDevServiceConfig.java
jfbenckhuijsen: Created #720

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseDevServiceConfig.java
jfbenckhuijsen: Config root has been updated as suggested. Lemme know what you think. Had to rework a bit to get it in line with the changes I did to the testcontainers module, on the bright side, the code is quite a bit cleaner now imho.

jfbenckhuijsen:
Ok, quite a bit of work so far:
- The code for the underlying FirebaseContainerEmulator has seen quite some changes. Most of them are to 

  1. support some additional features (like enable debugging, pipe logging from the container to the host, additional flags to setup the emulator etc
  2. Added a Builder API to create the configuration, which results in quite a bit of cleaner code than before, especially when translating between the Quarkus config and the emulator config
  3. The emulator classes are pulled to a separate package
  4. The loading logic for a custom firebase.json file has been greatly expanded, meaning the code will detect the majority of the settings there and use that. I.e. if you create a new firebase project and use your firebase.json file (with some minimal changes) you're good to go and can easily switch between the normal emulator (running via terminal or whatever) and the dockerized version. 

- ConfigRoot issues should be resolved now. Everything is neatly pulled into a "q.g.c.devserices.*" config root
- The creation code should automatically detect a firebase.json file in the root of the project (i.e. the current working directory for tests), so just including the module should be enough to get this running for any firebase project (the docs have been updated accordingly). 

The only issue I'm debugging right now (hence marked the branch as work in progress) is that the firebase-admin tests are failing. The issue is that the `io.quarkiverse.googlecloudservices.firebase.admin.runtime.authentication.http.FirebaseSecurityAuthMechanism` in the branch is not detected as a CDI bean for some reason. Haven't been able to figure out why. Injecting it from a test inside the module itself works fine, but for some reason I can't inject it from the firebase-auth integration test module

jfbenckhuijsen:
@loicmathieu question w.r.t. the currently failing build. Cause of that is that the tests try to validate that a function is running, which requires the packages needed for that function are installed (i.e. run `npm install`). As this hasn't been done in the build, it fails. Currently see the following options to resolve this:
1. Disable the unit test fully. Functions work, but we won't see any regressions
2. Run a npm install before running the maven build to get those dependencies installed
3. Move those tests to the integration tests and only run them manually. 
4. ...?

any preference?

docs/modules/ROOT/pages/firebase.adoc
loicmathieu: Thanks!

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainer.java
```diff
@@ -0,0 +1,1382 @@
...
+
+        /**
+         * Build the configuration
+         *
+         * @return The emulator configuration.
+         */
+        public EmulatorConfig buildConfig() {
+            if (firebaseConfig == null) {
+                // Try to autoload the firebase.json configuration
+                var defaultFirebaseJson = new File("firebase.json").getAbsoluteFile().toPath();
```
loicmathieu, line 416: This will not work in a JAR, you should use `getResourceAsStream()` and reads from META-INF

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainer.java
```diff
@@ -0,0 +1,1382 @@
...
+        }
+
+        private void setupFunctions() {
+            if (emulatorConfig.firebaseConfig().functionsConfig().functionsPath.isPresent()) {
+                this.dockerBuilder.run("mkdir -p " + containerFunctionsPath(emulatorConfig));
+                this.dockerBuilder.volume(containerFunctionsPath(emulatorConfig));
+            }
+        }
+
+        private void setupUserAndGroup() {
```
loicmathieu, line 1180: I'm not fan of creating user/group and chown, is it really necessary?

firebase-devservices/deployment/src/main/schema/firebase-config.json
loicmathieu, line 1: Please move it to META-INF

loicmathieu: But doesn't this schema exist somewhere?
I would rather use a distant schema so iy's always up-to-date or no schema and no validation.

firebase-devservices/deployment/src/test/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainerCustomConfigTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+@Testcontainers
+public class FirebaseEmulatorContainerCustomConfigTest {
+
+    private static final File tempEmulatorDataDir;
+
+    static {
+        try {
+            // Create a temporary directory for emulator data
+            tempEmulatorDataDir = Files.createTempDirectory("firebase-emulator-data").toFile();
```
loicmathieu, line 27: Better to create it somewhere it will be automatically deleted like in the target folder or /tmp

firebase-devservices/deployment/src/test/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainerIntegrationTest.java
```diff
@@ -0,0 +1,359 @@
...
+public class FirebaseEmulatorContainerIntegrationTest {
+
+    private static final File tempEmulatorDataDir;
+    private static final File tempHostingContentDir;
+
+    static {
+        try {
+            // Create a temporary directory for emulator data
+            tempEmulatorDataDir = Files.createTempDirectory("firebase-emulator-data").toFile();
+            tempHostingContentDir = Files.createTempDirectory("firebase-hosting-content").toFile();
```
loicmathieu, line 61: again, please create files somewhere it will be automatically deleted

firebase-devservices/deployment/storage.rules
loicmathieu, line 1: What is this file used for?

firebase-devservices/deployment/pom.xml
```diff
@@ -0,0 +1,158 @@
...
+
+                </configuration>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>generate</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
```
loicmathieu, line 154: What's the purpose of this? I don't think it's included by default.

loicmathieu (changes requested):
There is a lot of config/rules file, I'm not sure what it's used for, if we really need them, and if they should not be moved to META-INF so they are included in the JAR and loaded from there.

jfbenckhuijsen:
> If a simple `npm install` would make the test pass, let's do it! If it's more complex, move it to integration tests and runs it manually.

Its a simple npm install, I'll put that in the actions

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseDevServiceProjectConfig.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.Optional;
+
+import io.quarkus.runtime.annotations.ConfigRoot;
+import io.smallrye.config.ConfigMapping;
+
+/**
+ * Temporary Config root to retrieve the project id for the Firebase Emulator Container. We will remove this interface
+ * in the future in favour of using the common setup.
+ */
+@ConfigMapping(prefix = "quarkus.google.cloud")
```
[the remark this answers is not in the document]
jfbenckhuijsen: Gotcha, will switch and update documentation

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainer.java
jfbenckhuijsen: True, but that's actually intentional. In a firebase project (using the Firebase CI tools), a firebase.json file is generated at the root of the project itself (i.e. outside of the JAR). So by using this approach, you actually have the same behavior as the normal Firebase CI tools and the user won't see the difference

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainer.java
jfbenckhuijsen: Yes, that's what we discussed earlier in passing in the Github user/group, If you don't do this, docker will run with a different user (on CI builds at least, not for Docker Desktop), and you can;t access/read/remove/write the files from the host.

firebase-devservices/deployment/src/main/schema/firebase-config.json
jfbenckhuijsen: haha ,way ahead of ya ;). There is a schema for this, but there is an issue in that schema which prevents a correct generation of the classes from that. Referenced issue is https://github.com/firebase/firebase-tools/pull/7953 

I'm working on that too to get it merged into mainly, have a working version locally, but need to do some cleanup. Once that is merged, we can reference the external file.

The file itself is only used to generate some POJOs to ease reading the firebase.json file and mapping that config to the config of the plugin. I.e. there is no need to include that schema file at runtime. Given that, still prefer to have it in META-INF?

firebase-devservices/deployment/src/test/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainerCustomConfigTest.java
jfbenckhuijsen: true, on it

firebase-devservices/deployment/src/test/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainerIntegrationTest.java
jfbenckhuijsen: on it

firebase-devservices/deployment/storage.rules
jfbenckhuijsen: That's the configuration for Google Storage, which is used in one of the unit tests. It is included in the emulator docker image at runtime to configure the Storage emulator. I.e. just for testing purposes. 

In a normal firebase project, this is also the correct location for that file, so to validate that the stuff works in a regular project, I kept those files in the regular folders as much as possible.

firebase-devservices/deployment/pom.xml
jfbenckhuijsen: See above, that's to generate the POJOs to read the firebase.json config if the user provides one via the regular firebase CI tools. Based on the earlier mentioned schema, this file is read into the POJO's and those are used to configure the docker container.

jfbenckhuijsen:
> There is a lot of config/rules file, I'm not sure what it's used for, if we really need them, and if they should not be moved to META-INF so they are included in the JAR and loaded from there.

To summarize, those are all used for testing purposes, not used at runtime, at kept in those locations as this mimics the way a regular Firebase project would be setup, i.e. having those files there is part of the test setup.

firebase-devservices/deployment/src/main/schema/firebase-config.json
loicmathieu: Yes, all files should be in META-INF

loicmathieu:
> To summarize, those are all used for testing purposes, not used at runtime

OK, great, understood.
In this case, they must be inside the test directory, not anywhere else.

loicmathieu:
Everything that is test related should either be in src/test or in an integration-test module.

jfbenckhuijsen:
> If a simple `npm install` would make the test pass, let's do it! If it's more complex, move it to integration tests and runs it manually.

Fixed: 
https://github.com/quarkiverse/quarkus-google-cloud-services/pull/707/commits/f353ec02dccfeac7f078988a31912697fc6f559a
https://github.com/quarkiverse/quarkus-google-cloud-services/pull/707/commits/39608c812ccb1342ba98430c3c496cc679bf27f9

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseDevServiceProjectConfig.java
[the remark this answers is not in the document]
jfbenckhuijsen: Fixed https://github.com/quarkiverse/quarkus-google-cloud-services/pull/707/commits/bf90792a9331162b117b70bfc2225fedcc6a23ea

jfbenckhuijsen:
> Everything that is test related should either be in src/test or in an integration-test module.

Fixed: https://github.com/quarkiverse/quarkus-google-cloud-services/pull/707/commits/8be975cd5e09591084d38ce82ffc96178a7b69a2

firebase-devservices/deployment/src/test/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainerIntegrationTest.java
jfbenckhuijsen: Fixed https://github.com/quarkiverse/quarkus-google-cloud-services/pull/707/commits/338451a8c93df85c097d2ad005e9ccdf653c0a3a

firebase-devservices/deployment/src/test/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainerCustomConfigTest.java
jfbenckhuijsen: Fixed https://github.com/quarkiverse/quarkus-google-cloud-services/pull/707/commits/338451a8c93df85c097d2ad005e9ccdf653c0a3a

firebase-devservices/deployment/src/main/schema/firebase-config.json
jfbenckhuijsen: Fixed https://github.com/quarkiverse/quarkus-google-cloud-services/pull/707/commits/9c855f6f06bed61ceacfa7a5600308daa7bbec1f

jfbenckhuijsen:
All comments processed, your turn again ;)

loicmathieu (approved):
LGTM, thanks a lot!
This is a huge contribution.

I'll certainly do a round of cleanup later, if so, I'll send you the PR for validation.
I also didn't check the doc in details.
But overall, the PR is a nice addition so I'll merge it as is as it's already too old ;).

Thanks again for your patience

PR #711, opened 2024-11-15
Use NoCredentials in case the emulator host is being connected to

karrybit:
Dynamically use NoCredentials if the emulator is used.


loicmathieu (approved):
LGTM, thanks

loicmathieu:
Test are failing, something seems to didn't work well with the Storage emulator. Or maybe with how the current tests works.

jfbenckhuijsen:
@loicmathieu we can close this one

PR #728, opened 2025-01-03
Quarkus pub sub producers

jfbenckhuijsen:
@loicmathieu Cleaning up some old stuff for the new year. :) 

This PR adds the SubscriptionAdminClient and TopicAdminClient as CDI beans for some more advanced usage apart from the stuff offered by QuarkusPubSub. 

Don't know if we want this in there, if not lemme know and I'll throw it away.


loicmathieu:
I think it's a great addition (and should be documented somewhere).

But I'd rather add them in a standard factory class as we do for eg for Storage: https://github.com/quarkiverse/quarkus-google-cloud-services/blob/main/storage/runtime/src/main/java/io/quarkiverse/googlecloudservices/storage/runtime/StorageProducer.java

jfbenckhuijsen:
Moved to separate bean and docs added.

loicmathieu (approved):
LGTM, it made me learn a new trick today, first time I see the usage of `@Disposes` ;)

jfbenckhuijsen:
Haha, also had to look it up when building this code initially ;)

PR #729, opened 2025-01-03
Feature/firebase extra logging

jfbenckhuijsen:
These changes add some logging to help debug issues in the config or project setup.

Also a bug is fixed when using the automatically detected firebase.json file. In some cases this resulted in a container which cannot be started correctly.


loicmathieu (approved):
LGTM, thanks

PR #731, opened 2025-01-04
Feature/firebase experiments

jfbenckhuijsen:
Firebase has a way to enable experimental features. This is used e.g. when combining next.js with Firebase. The new option allows these experiments to be used.


jfbenckhuijsen:
@loicmathieu There is also some workaround code in this PR for a bug I ran into (which mostly triggers when using the experiments, hence I added it to this PR). Filed a bug (#732) to remove the logic once the upstream bug is fixed.

jfbenckhuijsen:
@loicmathieu Have some time to get this merged? Besides the feature, this PR also contains a few fixes for issues I ran in to while working with the extension quite a bit (i.e. really ironing out the corner cases). Can split the PR into multiple if that eases reviewing.

PR #772, opened 2025-02-18
Allow authentication based on session cookies

jfbenckhuijsen:
This features implements support for authentication as per the firebase documentation (https://firebase.google.com/docs/auth/admin/manage-cookies#java) 

The approach and integration into the extension was also discussed with some of the quarkus guys at (https://github.com/quarkusio/quarkus/discussions/46319)


firebase-admin/runtime/src/main/java/io/quarkiverse/googlecloudservices/firebase/admin/runtime/FirebaseAuthConfig.java
```diff
@@ -47,6 +47,61 @@ public interface AuthConfig {
...
+             */
+            Optional<String> name();
+
+            /**
+             * The expiration duration of the session cookie. Uses {@link java.time.Duration#parse(CharSequence)} to
+             * get a duration for the expiration. See the JavaDoc of this method for the format of this value.
+             * <p>
+             * Defaults to 5 days.
+             */
+            @WithDefault("P5")
```
loicmathieu, line 69:
```suggestion
            @WithDefault("P5D")
```

loicmathieu: Better to use a Duration and not a String

firebase-admin/runtime/src/main/java/io/quarkiverse/googlecloudservices/firebase/admin/runtime/FirebaseAuthConfig.java
```diff
@@ -47,6 +47,61 @@ public interface AuthConfig {
...
+             */
+            @WithDefault("false")
+            Optional<Boolean> validateToken();
+
+            /**
+             * Minimum token validity in case {@link #validateToken()} is set to true. Uses
+             * {@link java.time.Duration#parse(CharSequence)} to get a duration for the expiration. See the JavaDoc of
+             * this method for the format of this value.
+             */
+            Optional<String> minimumTokenValidity();
```
loicmathieu, line 89: You can directly use a Duration, there is a converter for it

firebase-admin/runtime/src/main/java/io/quarkiverse/googlecloudservices/firebase/admin/runtime/FirebaseAuthConfig.java
```diff
@@ -47,6 +47,61 @@ public interface AuthConfig {
...
+            /**
+             * Perform an additional check to see if the session was revoked
+             */
+            @WithDefault("true")
+            Boolean checkRevoked();
+
+            /**
+             * Validate the expiration date of the token.
+             */
+            @WithDefault("false")
```
loicmathieu, line 81: Is it aligned with what other security extensions are doing?

firebase-admin/runtime/src/main/java/io/quarkiverse/googlecloudservices/firebase/admin/runtime/FirebaseSessionCookieManager.java
```diff
@@ -0,0 +1,192 @@
...
+    boolean validateToken;
+
+    @ConfigProperty(name = "quarkus.google.cloud.firebase.auth.session-cookie.minimum-token-validity")
+    Optional<String> minimumTokenValidity;
+
+    @ConfigProperty(name = "quarkus.google.cloud.firebase.auth.session-cookie.login-api-path")
+    Optional<String> loginPath;
+
+    @ConfigProperty(name = "quarkus.google.cloud.firebase.auth.session-cookie.logout-api-path")
+    Optional<String> logoutPath;
```
loicmathieu, line 56: You can inject the configguration object itself, would be more easier instead of injecting each property individually.

firebase-admin/runtime/src/main/java/io/quarkiverse/googlecloudservices/firebase/admin/runtime/authentication/http/DefaultFirebaseIdentityProvider.java
```diff
@@ -67,11 +80,13 @@ public Uni<SecurityIdentity> authenticate(FirebaseAuthenticationRequest request,
...
      * @param token The FirebaseToken to be authenticated.
+     * @param rawToken The raw JWT token
      * @return A SecurityIdentity representing the authenticated user or null if authentication fails.
      */
-    public SecurityIdentity authenticate(FirebaseToken token) {
+    public SecurityIdentity authenticate(FirebaseToken token, String rawToken) {
         var builder = QuarkusSecurityIdentity.builder()
-                .setPrincipal(getPrincipal(token));
+                .setPrincipal(getPrincipal(token, rawToken))
+                .addCredential(new TokenCredential(rawToken, "Firebase"));
```
loicmathieu, line 89: As it's used somewhere else, better to use a constant to avoid duplicating in two different places the token credential type.

loicmathieu (changes requested):
There are a lof of modifications that seem unrelated in the documentation includes folder, are they included by error or missing artefact of a previous PR where documentation has been missing?

loicmathieu:
@sberyozkin as a Quarkus security maintainer can you have a look at this PR? Don't go into details on config or Google related changes but a security oriented review would be great.

jfbenckhuijsen:
> There are a lof of modifications that seem unrelated in the documentation includes folder, are they included by error or missing artefact of a previous PR where documentation has been missing?

I think those changes are caused by the recent quarkus upgrade as the changes all seem alike. Those dependabot bumps don't regenerate and commit the docs, so these are stale updates.

Can easily create a separate PR for this, or we include it, lemme know, either is fine,.

firebase-admin/runtime/src/main/java/io/quarkiverse/googlecloudservices/firebase/admin/runtime/FirebaseAuthConfig.java
jfbenckhuijsen: Ah cool, didn't know that, I'll update it.

firebase-admin/runtime/src/main/java/io/quarkiverse/googlecloudservices/firebase/admin/runtime/authentication/http/DefaultFirebaseIdentityProvider.java
jfbenckhuijsen: Fixed

firebase-admin/runtime/src/main/java/io/quarkiverse/googlecloudservices/firebase/admin/runtime/FirebaseSessionCookieManager.java
jfbenckhuijsen: Fixed

firebase-admin/runtime/src/main/java/io/quarkiverse/googlecloudservices/firebase/admin/runtime/FirebaseAuthConfig.java
jfbenckhuijsen: Fixed

firebase-admin/runtime/src/main/java/io/quarkiverse/googlecloudservices/firebase/admin/runtime/FirebaseAuthConfig.java
jfbenckhuijsen: Fixed

loicmathieu:
@jfbenckhuijsen I merged https://github.com/quarkiverse/quarkus-google-cloud-services/pull/784 can you rebase and remove the unnecessary changes?

jfbenckhuijsen:
> @jfbenckhuijsen I merged #784 can you rebase and remove the unnecessary changes?

Done

sberyozkin:
Apologies for missing out on it, @jfbenckhuijsen @loicmathieu.

I have to admit, it looks like to me this PR offers a new variation of the OIDC authorization code flow support in context of working with Firebase. 

I don't want to be saying `quarkus-oidc` is the only way to go, but this extension is integrated indirectly with `quarkus-oidc` and it will manage everything related to the code flow process, including the session cookies.

@jfbenckhuijsen Can you give me a favor please and explain what does it mean, support the authentication based on session cookies with Firebase, what is different when the session cookie based authentication is supported with the Google authentication. 

Please bear in mind that I haven't used Firebase - but I believe `quarkus-oidc` should be able to support user logins into Firebase.

Thanks

sberyozkin:
Thanks @jfbenckhuijsen 

>Now specifically this PR adds support for passing the JWT token between the client (normally the frontend app) and the backend app using session cookies instead of regular HTTP Auth Bearer headers in a secure way. 

Right, which is what `quarkus-oidc` can do too:

1. SPA gets the user login into a OIDC `web-app` application,
2. Session cookies are available. 
3. Session cookie is re-validated by Quarkus on every request (it's decrypted, token's signature is verified again, etc)

The flow you describe is:

1. SPA itself makes a login
2. SPA passes ID token to the server side endpoint which would return this token back to SPA. How is this session login method secured ? 
3. Now the session cookie is available to Quarkus. Can you clarify please how is this session cookie verified ?

Thanks

firebase-admin/runtime/src/main/java/io/quarkiverse/googlecloudservices/firebase/admin/runtime/authentication/http/FirebaseSecurityAuthMechanism.java
```diff
@@ -56,11 +67,18 @@ public Uni<SecurityIdentity> authenticate(RoutingContext context, IdentityProvid
...
+                    return identityProviderManager.authenticate(new FirebaseAuthenticationRequest(tokenChallenge, false));
                 }
             }
         }
 
+        if (sessionCookiesEnabled) {
+            var manager = cookieManagerInstance.get();
+            var cookie = manager.getSessionCookie(context);
+            return cookie.map(c -> identityProviderManager
+                    .authenticate(new FirebaseAuthenticationRequest(c, true))).orElse(Uni.createFrom().nullItem());
```
jfbenckhuijsen, line 79: @sberyozkin This is where the cookies are validated retrieved from the manager and authenticated using the identityProviderManager

firebase-admin/runtime/src/main/java/io/quarkiverse/googlecloudservices/firebase/admin/runtime/authentication/http/DefaultFirebaseIdentityProvider.java
```diff
@@ -48,11 +57,17 @@ public Class<FirebaseAuthenticationRequest> getRequestType() {
      */
     @Override
     public Uni<SecurityIdentity> authenticate(FirebaseAuthenticationRequest request, AuthenticationRequestContext context) {
+        Future<FirebaseToken> tokenFuture;
+        if (request.isCookie()) {
```
jfbenckhuijsen, line 61: @sberyozkin actual verification of the token in the cookie

firebase-admin/runtime/src/main/java/io/quarkiverse/googlecloudservices/firebase/admin/runtime/FirebaseAuthConfig.java
jfbenckhuijsen: btw @sberyozkin could you also let me know what you prefer here. Basically what this flag does:

- if false, the user will get a session token regardless of the remaining validity of the original JWT token
- if true, the user will only get a session token if his JWT token has a minimum validity

The default is currently false, as per the referenced google docs. Changing it is easy, but that would require we always specify a minimum validity (which I don't think we can set a default for). 

Agreed to leave it off by default and allow people with higher security requirements to specify this themselves?

sberyozkin:
Thanks @jfbenckhuijsen, I believe I can see how it is verified (it is forward to the Firebase to verify if I understand it correctly, which is good). 

The part which seems unusual, and I understand this is what Firebase docs suggest, is where SPA posts an `ID token` to the server to inform it that it is what this SPA wants to use for the authentication from now on, not the other way around, where the server initiates the creation of session cookies. 

So the session cookie is not created at the point where the actual authentication took place. I.e, the user logged in to SPA, not to Quarkus, and therefore SPA should only forward bearer access tokens to Quarkus to use it Google Cloud services, but with this solution the user has a session cookie with the server, this session cookie contains ID token. What about the access token which is needed to access the actual Google services ?

The implementation wise, it looks very neat, great job. 

However, given that the Firebase authentication mechanism already supports the bearer token authentication, and SPA can pass the bearer access token, I don't understand the goal which is meant to be achieved with adding this session cookie option.

As I said, it is a quality PR, thanks for it, I do not mean to make it difficult for this PR to go ahead, IMHO having more clarifications would help since I was asked to review.

Enjoy the weekend, @jfbenckhuijsen @loicmathieu, cheers !

jfbenckhuijsen:
@sberyozkin haha, no worries, happy with the help, security stuff is always hard, so the extra pair of eyes to get this done is appreciated.

As for your questions:
1. Basically the scenario covered with this setup is to protect your own application, although you can also forward the ID token from firebase to the google services (e.g Firestore works that way, and yes IIRC it uses ID token, not access token)
2. If you would want to use the google services using the access token, that's a different use case. I think you would probably still use the regular Auth-Bearer-Header approach (which is the default for the authorization btw, session cookies are only used when there is no header). Haven't tested that in practice though, so can't verify that easily
3. The main reason for adding session cookies is mostly to do with those situations where using a Auth header isn't possible (or convenient). The only case I know of (a stupid one, but one not with an easy solution), is if you want to serve images which are sensitive in nature to the user. You'd probably want to have those images behind a secured endpoint, so only authorized users can see them. Using a regular  tag won't work, as  doesn't set headers, which is where the cookies come in, as they are always set by the browser

HTH and enjoy the weekend :)

sberyozkin:
Thanks for the clarification, @jfbenckhuijsen.

While I think this mechanism is somewhat unusual, I can imagine that some users may prefer it.

But it occurred to me, while I was offline, that it is actually orthogonal to the work of this extension. I.e, this is about the user authentication into the Quarkus application, it does not facilitate the actual access to Google services supported by this extension. It appears to me it should be presented as a dedicated Quarkiverse extension which would offer Firebase session cookie based authentication to any random Quarkus application.

Did I get it wrong ? Thanks

jfbenckhuijsen:
@sberyozkin This is indeed about users authenticating into the Quarkus application. However, this is based on features/infrastructure provided by Google (Firebase). Firebase as a whole is an integrated platform to build applications, i.e. you can't separate that easily. For example, those users are also used in Firestore to do a sort of instance based security allowing users access to specific data (rows) in the database. Also, the session cookies are always based on the JWTs provided by Firebase Auth, so a separate extension would not be possible.

That said, would love if we could get this PR merged, so how can we bring this to a closure? And If needed continue this interesting discussion somewhere else ;)

sberyozkin:
@jfbenckhuijsen Well, this is orthogonal to the work of this extension IMHO. I'll let Loic who is this extension's maintainer decide. Thanks

jfbenckhuijsen:
> @jfbenckhuijsen Well, this is orthogonal to the work of this extension IMHO. I'll let Loic who is this extension's maintainer decide. Thanks

Awesome, with regards to the change. Does this imply you see no security related issues or needed changes (because that was the question Loic and me had 😀 )?

loicmathieu (approved):
As there are no security issue, I'll merge it, it adds a nice feature for Firebase application developer even if it's not specific to Google Cloud Services as it need the infrastructure provided here it's easier to integrate it here.

Issue #586, opened 2024-02-28
Test suite takes a long time to shut down


jamesnetherton:
Not sure what's going on here, but it's taking 6+ minutes for the application under test to stop.

```
INFO  [io.quarkus] (main) quarkus-google-cloud-services-main-it stopped in 360.024s
```

loicmathieu:
Juts noticed that, it seems to be related to Cloud Logging.

jfbenckhuijsen:
@loicmathieu We can close this based on merging #777

loicmathieu:
Yes, fixed by https://github.com/quarkiverse/quarkus-google-cloud-services/pull/777

By the way, better to include the issue number in the PR, I didn't remember all open issues ;) 
I usually even use Github special syntax to close them automatically.

PR #777, opened 2025-02-21
Add dispose methods for resources so close is called

jfbenckhuijsen:
Reduces shutdown time for the main IT

### DIFF
--- a/pubsub/runtime/src/main/java/io/quarkiverse/googlecloudservices/pubsub/QuarkusPubSub.java
+++ b/pubsub/runtime/src/main/java/io/quarkiverse/googlecloudservices/pubsub/QuarkusPubSub.java
@@ -5,7 +5,6 @@ import java.util.Optional;
 import java.util.stream.StreamSupport;
 
 import jakarta.annotation.PostConstruct;
-import jakarta.annotation.PreDestroy;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.inject.Instance;
 import jakarta.inject.Inject;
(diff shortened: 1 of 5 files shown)


jfbenckhuijsen:
Fixes #586

pubsub/runtime/src/main/java/io/quarkiverse/googlecloudservices/pubsub/QuarkusPubSub.java
```diff
@@ -46,13 +45,6 @@ void init() {
         }
     }
 
-    @PreDestroy
-    void destroy() throws Exception {
-        if (channelProvider.isPresent()) {
-            channelProvider.get().getTransportChannel().close();
-        }
-    }
```
loicmathieu, line 54: Why do you remove it?
As far as I rember, without this, ITs shows error messages...

jfbenckhuijsen: I'll have a look at it, but the main reason is that this causes the slow down in the main IT we see. Basically, that channel provider and the transport channel is owned by the pubsub logic itself, so we shouldn't close it manually. It will get automatically closed at the appropriate time by the pubsub logic.

loicmathieu: OK then +1

loicmathieu (approved):
LGTM

jfbenckhuijsen:
I'll try to have a look at those error message somewhere in the future

PR #781, opened 2025-02-25
Add dev service for Spanner

### LINKED ISSUE #400: Spanner does not work against emulator host
The current state of the Google Cloud Spanner Quarkus extension does not work at all using the Spanner emulator. The reason is actually pretty simple: it requires valid credentials and there is no way around it, even the [documented approach](https://quarkiverse.github.io/quarkiverse-docs/quarkus-google-cloud-services/main/index.html#emulators) does not work.

A way to make it work properly with an emulator is pretty simple, and does _not_ require the hack from the docs described in the link above.

The [current code](https://github.com/quarkiverse/quarkus-google-cloud-services/blob/1.4.0/spanner/runtime/src/main/java/io/quarkiverse/googlecloudservices/spanner/runtime/SpannerProducer.java) is pretty straight forward, but makes using an emulator practically impossible, because it has a hard dependency via `@Inject GoogleCredentials googleCredentials;`.

A way around that and the hack is to programmatically request the `GoogleCredentialy` and only bark, it the emulator host is _not_ present, like in this piece of code:
```java
@ApplicationScope
public class SpannerProducer {

    @Inject
    GcpConfigHolder gcpConfigHolder;

    @Inject
    SpannerConfiguration spannerConfiguration;

    @Produces
    @Singleton
    @Default
    public Spanner storage() throws IOException {
        GoogleCredentials googleCredentials = null;
        Exception googleCredentialsException = null;
        try {
            googleCredentials = Arc.container().instance(GoogleCredentials.class).get();
        } catch (Exception e) {
            googleCredentialsException = e;
        }
    
        GcpBootstrapConfiguration gcpConfiguration = gcpConfigHolder.getBootstrapConfig();
        SpannerOptions.Builder builder =
            SpannerOptions.newBuilder().setProjectId(gcpConfiguration.projectId.orElse(null));
        if (!spannerConfiguration.emulatorHost().isPresent()) {
            builder.setEmulatorHost(spannerConfiguration.emulatorHost().get());
        } else {
            if (googleCredentials == null) {
                if (googleCredentialsException != null) {
                    throw new RuntimeException("No GoogleCredentials available", googleCredentialsException);
                } else {
                    throw new RuntimeException("No GoogleCredentials available");
                }
            }
            builder.setCredentials(googleCredentials);
        }

        return builder.build().getService();
    }
```

### LINKED ISSUE #542: Provides Spanner dev services
Provides dev services support for Google Cloud Spanner

jfbenckhuijsen:
closes #542 
closes #400

### DIFF
--- a/spanner/deployment/src/main/java/io/quarkiverse/googlecloudservices/spanner/deployment/SpannerDevServiceConfig.java
+++ b/spanner/deployment/src/main/java/io/quarkiverse/googlecloudservices/spanner/deployment/SpannerDevServiceConfig.java
@@ -15,7 +15,7 @@ import io.smallrye.config.WithDefault;
  *
  * <pre>
  * quarkus.google.cloud.spanner.devservice.enabled = true
- * quarkus.google.cloud.spanner.devservice.image-name = gcr.io/cloud-spanner-emulator/emulator:1.4.0 # optional
+ * quarkus.google.cloud.spanner.devservice.image-name = gcr.io/cloud-spanner-emulator/emulator:1.5.9 # optional
  * quarkus.google.cloud.spanner.devservice.emulatorPort = 8085 # optional
  * </pre>
  */
@@ -33,9 +33,9 @@ public interface SpannerDevServiceConfig {
     /**
      * Sets the Docker image name for the Google Cloud SDK.
      * This image is used to emulate the Spanner service in the development environment.
-     * The default value is 'gcr.io/cloud-spanner-emulator/emulator:1.5.29'.
+     * The default value is 'gcr.io/cloud-spanner-emulator/emulator'.
      */
-    @WithDefault("gcr.io/cloud-spanner-emulator/emulator:1.5.29")
+    @WithDefault("gcr.io/cloud-spanner-emulator/emulator")
     String imageName();
 
     /**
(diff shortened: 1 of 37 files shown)


spanner/deployment/src/main/java/io/quarkiverse/googlecloudservices/spanner/deployment/SpannerDevServiceConfig.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    @WithDefault("false")
+    boolean enabled();
+
+    /**
+     * Sets the Docker image name for the Google Cloud SDK.
+     * This image is used to emulate the Spanner service in the development environment.
+     * The default value is 'gcr.io/cloud-spanner-emulator/emulator:1.5.29'.
+     */
+    @WithDefault("gcr.io/cloud-spanner-emulator/emulator:1.5.29")
```
loicmathieu, line 38: Wouldn't it be better to use latest as the default tag? (so no tag)

loicmathieu (changes requested):

spanner/deployment/src/main/java/io/quarkiverse/googlecloudservices/spanner/deployment/SpannerDevServiceConfig.java
jfbenckhuijsen: Pro: easier to stay up to date
Con: builds may be less stable if there are unforeseen changes to the emulator image (although it might be nice to see that)

Latest is used in other places (except the firebase base Node version, but that's because firebase-tools has restrictions on the engine version, so latest won't work there).

Changing it to be more consistent with the other setups

loicmathieu (approved):
LGTM, thanks

loicmathieu:
Tests didn't pass, can you have a look?

PR #797, opened 2025-03-14
Fixed issue with running as a group already existing in the docker image

jfbenckhuijsen:
There was an issue when an already existing group was used in the docker image. The logic tried to re-add the existing group, which of course failed the docker build. This especially manifested after the recent auto-uid-gid change. With this setup, this is fixed, as the logic first checks if the user exists.

Also, I extracted creation of the base image to a Dockerfile. This reduces the Java logic and is more easier to maintain, as it is now just a regular Dockerfile with some build arguments.

Moving the rest of the logic to a Dockerfile didn't seem the way to go, as this is way less straightforward in it logic (way more optional paths).


loicmathieu (approved):
LGTM, this is more declarative by using a Dockerfile and I think more maintainable.

Thanks.

PR #804, opened 2025-03-21
Fix reading the Dockerfile resources from a JAR

jfbenckhuijsen:
In case the extension was used as a JAR file, the resource could not correctly be read from the JAR file, as the ZipFileSystem was not automatically created (see https://docs.oracle.com/javase/7/docs/technotes/guides/io/fsp/zipfilesystemprovider.html ). This provides a fix for reading the file.


loicmathieu (approved):
LGTM thanks

PR #806, opened 2025-03-24
Automatically use NoCredentials in case the emulator host is being connected to.

jfbenckhuijsen:
Fixed/reimplementation of #711


loicmathieu (approved):
LGTM, thanks

PR #810, opened 2025-03-31
Don't close the filesystem as this will cause an error when reading the Dockerfile itself later on.

jfbenckhuijsen:


loicmathieu (approved):
LGTM, thanks

PR #811, opened 2025-03-31
Different strategy for building the base image

jfbenckhuijsen:
- Needed files are extracted from the JAR and moved to a temporary directory
- The docker image is build from the files in the temp dir.

### DIFF
--- a/firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainer.java
+++ b/firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainer.java
@@ -2,10 +2,7 @@ package io.quarkiverse.googlecloudservices.firebase.deployment.testcontainers;
 
 import java.io.File;
 import java.io.IOException;
-import java.net.URISyntaxException;
-import java.nio.file.FileSystemNotFoundException;
-import java.nio.file.FileSystems;
-import java.nio.file.Path;
+import java.nio.file.*;
 import java.util.*;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainer.java
```diff
@@ -1101,58 +1101,66 @@ public ImageFromDockerfile build() {
         }
 
         private void buildBaseImageIfNeeded() {
+            LOGGER.info("Checking if we need to build a new base image for the Firebase dev service image");
```
loicmathieu, line 1104:
```suggestion
            LOGGER.debug("Checking if we need to build a new base image for the Firebase dev service image");
```

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainer.java
```diff
@@ -1101,58 +1101,66 @@ public ImageFromDockerfile build() {
...
-                        .withBuildArg("GROUP_ID", "" + dockerGroup())
+                        .withBuildArg("BASE_IMAGE", baseImageName)
+                        .withBuildArg("USER_ID", "" + user)
+                        .withBuildArg("GROUP_ID", "" + group)
                         .withBuildArg("FIREBASE_VERSION", firebaseVersion);
 
                 baseImage.get();
                 this.dockerBuilder = null;
+            } else {
+                LOGGER.info("Base image already exists. Using the cached version");
```
loicmathieu, line 1127:
```suggestion
                LOGGER.debug("Base image already exists. Using the cached version");
```

If you think it deserves a log, you should emit a single log with information in it, here there is 2 logs when the base image already exist and one otherwise.

loicmathieu (changes requested):

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainer.java
jfbenckhuijsen: Fixed

firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/testcontainers/FirebaseEmulatorContainer.java
jfbenckhuijsen: Fixed, added new log line when there is no image. Changed level. 

Squashed and force pushed

loicmathieu (approved):
LGTM, thanks

PR #825, opened 2025-05-06
Add methods to customize the producer being created.

jfbenckhuijsen:
This allows for more advanced settings to be configured like message ordering, etc while still connecting correctly to the emulator.


loicmathieu (approved):
LGTM

PR #851, opened 2025-06-30
Add a link to the article

loicmathieu:


jfbenckhuijsen:
@loicmathieu just saw this issue yesterday. Seems to be related to the new Firebase tools version. If you’re in a hurry, downgrading to the previous version should fix this. 

I’ll have a look at the reason why this is happening in the meantime

PR #853, opened 2025-07-01
Fix setup of emulators.

jfbenckhuijsen:
It seems the new setup of firebase-tools now writes to the current directory. By default, this is the root directory, for which the running user does not have permissions. By settings the workdir to /srv/firebase, downloading emulators should work.


jfbenckhuijsen:
@loicmathieu issues seems fixed with this PR

PR #867, opened 2025-07-23
Fix setup of emulators.

jfbenckhuijsen:
Java 21 seems to be needed from now on.


jfbenckhuijsen:
@loicmathieu do we want to merge these changes?

Basically what this does is read a package.json file where the version from firebase-tools is specified and use that as the version to install in the docker image. 

This should allow something like dependabot or renovate to pick it up and just create branches for updating the version, instead of breaking the build because we use "latest".

PR #869, opened 2025-07-24
Change to the emulators version, which supports java 21, needed for the emulators.

jfbenckhuijsen:


loicmathieu (approved):
LGTM, thanks.

PR #879, opened 2025-08-15
Add some documentation how to use dev services on Bitbucket

jfbenckhuijsen:
Ok, weird stuff, had a running build with the previous solution, build then it stopped working.

Had a look at it with Atlassian support and this should be the correct way to do it. Multiple builds are stable now.


loicmathieu (approved):
Lgmt thanks

Issue #901, opened 2025-10-11
PubSub without requiring GOOGLE_APPLICATION_CREDENTIALS


Fungrim:
Hi, currently Pub/Sub requires `GOOGLE_APPLICATION_CREDENTIALS` which is a bit limiting. For example, I'm running a lot of code on a local server using docker compose, but I don't want to write my service account JSON to a file. 

I have a patch I can do a pull request on. It's a simple 5-liner that allows you to provide alternative `GoogleCredential` instances and will pick the one with the highest priority. So to solve my issue I used the patch and then...

```
@ApplicationScoped
public class PubSubAuthProvider {

    @ConfigProperty(name = "quarkus.google.cloud.service-account-encoded-key")
    String base64Key;

    @Produces
    @Singleton
    @Priority(Integer.MAX_VALUE)
    public CredentialsProvider credentialsProvider() throws IOException {
        final InputStream keyStream = new ByteArrayInputStream(Base64.getDecoder().decode(base64Key));
        final GoogleCredentials credentials = GoogleCredentials.fromStream(keyStream).createScoped("null");
        return () -> credentials;
    }
}
```

... which worked nicely. 

The better solution would be to harmonized Pub/Sub to directly allow `quarkus.google.cloud.service-account-encoded-key` like the other extensions - but I had limited time and the provider patch was simple :-)

Let me know if you want to see a pull request.

jfbenckhuijsen:
@loicmathieu I think we can close this?

loicmathieu:
@Fungrim can we close this one?

PR #903, opened 2025-10-13
Making credential provider instances be selected by priority

Fungrim:
Re-opening of PR for #901 - now with correct documentation.


loicmathieu (commented):
As the credential provider is annotated with `@Default` you should already be able to add another implementation that should replace this one.

I need to have a look, there should be an easier way to do that...

loicmathieu:
@Fungrim please test it before we proceed with this PR.

Fungrim:
`@Alternative` and `@Priority` works. So I'll close this PR.

PR #926, opened 2025-12-01
Upgrade to Quarksu 3.30

loicmathieu:


loicmathieu:
@mswiderski yes, definitly, I was going to update to 3.30 and the latest Google Cloud Library at the same time which seems to be challenging.
I'll try to only set the native configuration you give me, it should at least allow to upgrade to Quarkus 3.30.

mswiderski:
Excellent, thanks any plans to have a reales with quarkus 3.30.x ?

PR #944, opened 2026-01-02
Split push logic into a separate extension to resolve dependency issues when using the module combined with google cloud functions

### LINKED ISSUE #910: PubSub dependency does not work with latest Quarkus Google Cloud Functions LTS 3.27.0
In our project we are using Quarkus Google Cloud Functions and some modules from this project like `quarkus-google-cloud-logging` and `quarkus-google-cloud-pubsub` and this whole setup works perfectly fine with Quarkus Google Cloud Functions v3.20.0 and `quarkus-google-cloud-services` v2.15.0. 

However after upgrading to latest LTS version of quarkus 3.27.0, the integration tests and also the functionality inside GCP of this cloud function no longer works. I was able to pinpoint the issue to `quarkus-google-cloud-pubsub` module. It makes the integration test return 404 instead of processing events. 

I am providing here a minimum reproducible sample project. Here even though PubSub is not used anywhere in the project if I add `quarkus-google-cloud-pubsub` dependency in pom.xml the integration test no longer works. If I comment out the `quarkus-google-cloud-pubsub` dependency then the integration test works perfectly fine.

[google-cloud-functions.zip](https://github.com/user-attachments/files/23362170/google-cloud-functions.zip)

With other module like `quarkus-google-cloud-logging` there is no issue if its added to the project. For `quarkus-google-cloud-pubsub`  I've tried with different version and none of them work with Quarkus 3.27.0

Thank you for your contributions and help. I really appreciate all the effort that is put into this project.

jfbenckhuijsen:
fixes #910


jfbenckhuijsen:
N.b. will squash once review is done

PR #945, opened 2026-01-02
Removed log file which shouldn't be committed

jfbenckhuijsen:
Added gitignore file to prevent future commits


loicmathieu (approved):
LGTM, thanks

PR #951, opened 2026-01-07
Update to Google Cloud Library 26.71.0

loicmathieu:
26.73.0 causes a dependency issue with Firebase so we had to only upgrade to 26.71.0.

There were multiple native-related issues:
- A missing class from the ZSTD lib so we add this lib for native image compilation.
- Arrow Netty classes changed packages
- Spanner init early Netty shaded libraries: we switch to use Netty shaded instead of Quarkus provided one

Closes #936


jfbenckhuijsen:
Correction, we need to use the netty versions probably, see my PR #962

PR #952, opened 2026-01-07
Compose the trace string of the default project id if present.

jfbenckhuijsen:
This implementation aligns with the documentation of the config property. The old implementation disregarded the default project config.


loicmathieu (approved):
LGTM, thanks.
However, CI didn't pass

PR #954, opened 2026-01-07
Extended the documentation of default-log

### LINKED ISSUE #769: Provide beter documentation for required logging config value
The documentation around the required property `quarkus.google.cloud.logging.default-log` leaves a lot to be desired. It's not clear what this should actually be set to, or what it is referencing. Trying to google what "Google Operations log" might mean is a dead end. Is this the "log name" to be used? Some example values or a better description of this would go a long way to making this integration easier set up.

Also, the 'basic' config section is a bit confusing where it says;

> Currently, the logging system uses a single Google Cloud Operations log and this is a mandatory configuration parameter.

but lists two parameters;

* quarkus.google.cloud.logging.default-log
* quarkus.google.cloud.project-id

Are both required? It doesn't seem to be the case as I can omit project-id and it does't complain, but I also don't know what it does when I omit it.

**edit;** I'm assuming I am mis-interpreting/configuring this as when I include the logging ext, and enable it I'm getting a permission denied error, but the none gcp-formatted logs are still showing up fine. My current config is just;

```
quarkus.google.cloud.logging.format=JSON
# ????
quarkus.google.cloud.logging.default-log=stdout
```
`ERROR: onFailure exception: com.google.cloud.logging.LoggingException: io.grpc.StatusRuntimeException: PERMISSION_DENIED: Permission 'logging.logEntries.create' denied on resource (or it may not exist).`

jfbenckhuijsen:
The original documentation was not really clear on which field on GCP was configured and it's limitations.

fixes #769

### DIFF
--- a/docs/modules/ROOT/pages/logging.adoc
+++ b/docs/modules/ROOT/pages/logging.adoc
@@ -73,7 +73,7 @@ public class LoggingResource {
 === Basic
 Currently, the logging system uses a single Google Cloud Operations log and this is a mandatory configuration parameter.
 
-* `quarkus.google.cloud.logging.default-log` - This is the nam of the resource to which the log belogs. See
+* `quarkus.google.cloud.logging.default-log` - This is the name of the resource to which the log belongs. See
   https://docs.cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry[the google docs] for more info. Note that
   currently only a single log id is supported. The logName will always be a project logName and this property does not
   accept forward-slashes.


docs/modules/ROOT/pages/logging.adoc
```diff
@@ -73,8 +73,11 @@ public class LoggingResource {
 === Basic
 Currently, the logging system uses a single Google Cloud Operations log and this is a mandatory configuration parameter.
 
-* `quarkus.google.cloud.logging.default-log`
-* `quarkus.google.cloud.project-id`
+* `quarkus.google.cloud.logging.default-log` - This is the nam of the resource to which the log belogs. See
```
loicmathieu:
```suggestion
* `quarkus.google.cloud.logging.default-log` - This is the name of the resource to which the log belongs. See
```

loicmathieu (approved):

PR #1061, opened 2026-06-25
Keep emulator host ports reachable on shared network

jfbenckhuijsen:
When quarkus.devservices.launch-on-shared-network is enabled, the Firebase emulator container previously skipped exposing/binding ports to the Docker host entirely and switched JVM-facing config properties to the shared-network alias (<alias>:<internalPort>). Since the Quarkus JVM process is not part of the Docker network, this made the emulators unreachable from the application itself.

- FirebaseEmulatorContainer now always exposes/binds emulator ports to the host, regardless of shared-network mode, while still joining the shared network and keeping its network alias for in-network container consumers (emulatorUrl/emulatorUrls).
- Added hostMappedPort/hostEmulatorUrl/hostEmulatorUrls, and changed emulatorEndpoints() (which feeds the JVM-facing CONFIG_PROPERTIES) to use these host-mapped, localhost-based endpoints unconditionally. Extracted a shared withHttpPrefixIfNeeded() helper to remove duplicated http:// prefix logic between the host- and container-facing URL builders.
- FirebaseDevServiceProcessor now additionally, when on a shared network, calls Testcontainers.exposeHostPorts() for each emulator's host-mapped port and exposes new "container-emulator-host"/"container-host-override" config properties pointing at host.testcontainers.internal:<port>.

### DIFF
--- a/firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseDevServiceProcessor.java
+++ b/firebase-devservices/deployment/src/main/java/io/quarkiverse/googlecloudservices/firebase/deployment/FirebaseDevServiceProcessor.java
@@ -5,6 +5,7 @@ import java.util.*;
 import java.util.stream.Collectors;
 
 import org.jboss.logging.Logger;
+import org.testcontainers.Testcontainers;
 
 import io.quarkiverse.googlecloudservices.firebase.deployment.testcontainers.FirebaseEmulatorContainer;
 import io.quarkus.deployment.IsNormal;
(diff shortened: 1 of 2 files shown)


jfbenckhuijsen:
Note: I found these changes were needed when working on the playwright extension to get it to run the browser in a docker container. In this case, the playwright browser needs to access both the host (for e.g. REST calls to the Quarry's app) and the emulators for the browser. 

From the docker container, you can reach this if using the host.testcontaienrs.internal hostname if running on a shared network, but this conflicted with the setup of the shared networking code.

loicmathieu (approved):
LGTM, thanks

PR #1063, opened 2026-06-29
Avoid returning raw FirebaseOptions from REST test resource

jfbenckhuijsen:
Quarkus 3.37 changed how RESTEasy Reactive's Jackson integration generates serializers for response types: instead of invoking getters reflectively, it now calls them directly via generated bytecode. FirebaseOptions has a package-private getFirestoreOptions() getter, which RESTEasy Reactive's serializer can no longer access once the resource returns the raw SDK type, causing an IllegalAccessError at runtime.

FirebaseAppResource only ever needs to expose the project id for its tests, so return a small FirebaseOptionsDto(projectId) instead of the full FirebaseOptions object. This avoids the access error and stops leaking the whole SDK options object (including credentials) over HTTP.

Fixes failing build on #1056


loicmathieu (approved):
LGTM

PR #1067, opened 2026-07-01
Fixed the way pubsub config values are passed to the recorder

jfbenckhuijsen:
Previous, this setup works (although it was not correct). Since Quarkus 3.7, the core has become more strict and runtime values are no longer resolved in a recorder when a Build-time config is used. In the new setup, the audience, etc values are correctly read at runtime instead of resolved at buildtime. As the audience is most often configured via environment parameters (from the Pub Sub config on GCP) this setup is needed to get Pub sub working again.


jfbenckhuijsen:
@loicmathieu we seem to have a regression in Quarkus 3.7 combined with the pubs module. Think this should fix it, we might want to do a bug fix release?

loicmathieu (approved):
LGTM

PR #1085, opened 2026-07-27
Add support for Vite

jfbenckhuijsen:
Vite allows hot reload of the web code, however this was blocked in the way Vite was exposed through the firebase emulator in the docker container. Logic has been added to expose vite correctly (including documentation on the needed configuration changes).

In the future, support for more web frameworks may be added (same as firebase itself supports), so a strategy/factory approach was used here.


loicmathieu (approved):
LGTM, thanks

PR #1086, opened 2026-08-02
Improve support for vite with configurable port and improve networking

jfbenckhuijsen:
The vite port is configurable now, as this was the only case where running two services side-by-side could result in errors. Also clean up the networking setup and document it, so it is more consise and the various situations are documented correctly.

Also, this fixes an issue when running in a containerized integration test and the urls for the emulators were not set correctly.


loicmathieu (approved):
LGTM, thanks

PR #1088, opened 2026-08-07
Dependabot/gc 26.86.0 fix

jfbenckhuijsen:
Google HTTP Client is used in version 1.4.7 in Quarkus, whereas Google requires the 2.2.0 version. The 2.2.0 has been added to our parent pom, because otherwise the Quarkus BOM import takes precedence.

Also fixed a dependency convergence between spanner/bigtable/storage for an otel contrib dependency.

@loicmathieu this results in a clean build locally, however I'm a bit hesitant for this fix, is this is a major bump in the http-client version on the quarkus side. I don't have full overview where that dependency is used within the quarkus ecosystem, so no idea what will break for other extensions or core quarkus if we perform this upgrade.


loicmathieu:
So now there is a native issue :( 

But yes, having an HTTP library not aligned with the Quarkus core might be an issue for other Google Cloud related extentions like functions/OTEL/... so it cannot be bumped like that.

If it's what cause the issue, we need to delay the library bump until Quarkus upgrades to the HTTP client 2.

PR #1091, opened 2026-08-11
Shim to fix dependency divergence between Firestore and Quarkus

jfbenckhuijsen:
Google Cloud Firestore uses and older version of the otel libraries. These libraries have divergde in API, while still keeping a minor version upgrade. Quarkus new version of the library wins in the maven setup, resulting in NoSuchMethodErrors because firestore tries to use the now non-existing API. This shim rewires the offending part of Google Firestore to use the new API.


loicmathieu (approved):
LGTM

PR #1094, opened 2026-08-13
Expose hub port as a property

jfbenckhuijsen:
The hub can be used to do some advanced features via its rest interface. Expose it so we can use that.


loicmathieu (approved):
LGTM, thanks

PR #1096, opened 2026-08-16
Check if GRPC native support is sufficient

jfbenckhuijsen:
Check what happens if we just remove the GRPC support, maybe GAX has already implemented support by itself


jfbenckhuijsen:
@loicmathieu can you have a look at this? Was looking at the MTLTS support for GAX-grpc, and just tried this for fun, but is seems everything runs fine without these graal configuration. Can we just ditch this? I'm not sure if our IT tests are sufficient in terms of coverage of this.

loicmathieu (approved):
LGTM

loicmathieu:
Yes, Google Lib native support seems to be enough, let's remove this.

PR #1102, opened 2026-08-17
Fix native setup for FirebaseAdmin

jfbenckhuijsen:
Usage of FirebaseAdmin fails in a native environment. These changes should fix that. Note that not all uses have been validated, so we may need other changes than these, but its a start.

### DIFF
--- a/integration-tests/firebase-admin/src/test/java/io/quarkiverse/googlecloudservices/it/firebaseadmin/FirebaseAuthTest.java
+++ b/integration-tests/firebase-admin/src/test/java/io/quarkiverse/googlecloudservices/it/firebaseadmin/FirebaseAuthTest.java
@@ -2,19 +2,27 @@ package io.quarkiverse.googlecloudservices.it.firebaseadmin;
 
 import static io.restassured.RestAssured.given;
 
-import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.eclipse.microprofile.config.ConfigProvider;
 import org.junit.jupiter.api.BeforeEach;
 
+/**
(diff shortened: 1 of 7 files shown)


integration-tests/firebase-admin/src/test/java/io/quarkiverse/googlecloudservices/it/firebaseadmin/FirebaseAuthTest.java
```diff
@@ -2,19 +2,27 @@
...
+ * <p>
+ * {@link ConfigProvider} is used to retreive the config valeus over injection as this also needs to run in a integration
+ * test.
+ */
 public abstract class FirebaseAuthTest {
 
-    @ConfigProperty(name = "quarkus.google.cloud.project-id")
     String projectId;
 
-    @ConfigProperty(name = "quarkus.google.cloud.firebase.auth.emulator-host")
```
loicmathieu, line 13: Why can't we rely on `@ConfigProperty`?

loicmathieu (approved):
LGTM but before merging can you explain why you switch out of `@ConfigProperty` ?

integration-tests/firebase-admin/src/test/java/io/quarkiverse/googlecloudservices/it/firebaseadmin/FirebaseAuthTest.java
jfbenckhuijsen: Its the base class for an integration test`@QuarkusIntegrationTest` AFAIK ConfigProperty doesn't work there

PR #1104, opened 2026-08-20
Use ScheduledExecutorService provided by Quarkus for GRPC calls in Firestore

jfbenckhuijsen:
Running into some weird issues where thread context (e.g. request scope) is not propagated when using Firestore. I think using a quarkus / cdi managed executor should fix that.


loicmathieu (approved):
LGTM

PR #1108, opened 2026-08-24
Add a config property to disable push authentication.

jfbenckhuijsen:
PubSub emulator does not support authentication, so in a test scenario when using push, we need to disable the auth.


loicmathieu (approved):
LGTM

PR #1114, opened 2026-08-26
Fix index google-cloud-firestore and this extension's own runtime jar for quarkus-shim

jfbenckhuijsen:
Regression: the missing Jandex since 0.4.0 shim resulted in the Shim not being applied. Added a fix for this and a test to verify the behaviour

Added test to verify shim is applied
Added indexing so shim is applied, since 0.4.0 this didn't work (silently), the regression test should no signal this.

### DIFF
--- a/firestore/deployment/src/main/java/io/quarkiverse/googlecloudservices/firestore/deployment/FirestoreBuildSteps.java
+++ b/firestore/deployment/src/main/java/io/quarkiverse/googlecloudservices/firestore/deployment/FirestoreBuildSteps.java
@@ -4,6 +4,7 @@ import io.quarkiverse.googlecloudservices.firestore.runtime.FirestoreProducer;
 import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
+import io.quarkus.deployment.builditem.IndexDependencyBuildItem;
 
 public class FirestoreBuildSteps {
     protected static final String FEATURE = "google-cloud-firestore";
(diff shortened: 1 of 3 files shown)


firestore/deployment/src/main/java/io/quarkiverse/googlecloudservices/firestore/deployment/FirestoreBuildSteps.java
```diff
@@ -17,4 +18,23 @@ public FeatureBuildItem feature() {
...
+    }
+
+    /**
+     * Make sure that the runtime jar of this extension is indexed, so that {@code FirestoreProducer} can be found by
+     * {@code @Shim}
+     */
+    @BuildStep
+    public IndexDependencyBuildItem indexOwnRuntimeJarForShim() {
+        return new IndexDependencyBuildItem("io.quarkiverse.googlecloudservices", "quarkus-google-cloud-firestore");
+    }
```
loicmathieu: This should not be needed

loicmathieu (changes requested):

firestore/deployment/src/main/java/io/quarkiverse/googlecloudservices/firestore/deployment/FirestoreBuildSteps.java
jfbenckhuijsen: Build (and shim application) fails without this, reverted removal

loicmathieu: This is fishy, you index the library on the very library build step which is a kind of cycle indexation.
I think the issue is somewhere else on the build, it should miss a dependency.

Where does it fail and with which stacktrace?

jfbenckhuijsen: Basically what is does, is not index, i.e. no stack trace, but them shim doesn't get applied

loicmathieu (approved):
LGTM, too bad we had to do that

jfbenckhuijsen:
Agreed. We have a tracking bug for that, but I don't see much changes on the Firestore side, sadly...

PR #1115, opened 2026-08-26
Implement support for end-to-end testing using pubsub and push

jfbenckhuijsen:
Also adds some convenience methods to create topics and subscriptions for both the pull and push model.

Register host.docker.internal for Linux environments, so the pubsub emulator can access the host machine (when not running in shared network mode)


loicmathieu (approved):
LGTM, thanks

PR #1138, opened 2026-09-22
Add reflection for PubSubMessageJson class to allow for native compilation using push messages.

jfbenckhuijsen:
This class gets deserialized during message receival, so needs to be marked for reflection by graal.


jfbenckhuijsen:
@loicmathieu can you do a check if I set all the correct flags for serialization? I'm not fully into this part.

